PR #7, opened 2014-11-28
Expose layer to the current request.

izaakschroeder:
Sometimes you wish to know something about the layer that is currently executing which cannot be provided by `req.route`, particularly because `req.route` is not available for the `.use()` family _and_ because the route object doesn't say which member of its stack it is currently executing.

Example usage might be something like:

``` javascript
router.use(function middleware(req, res, next) {
    req.layer.handle === middleware; // true
    console.log(req.layer.name);
    next();
});
```

This change is non-breaking unless some existing middleware sets a `.layer` property on the `req` object in which case this will overwrite it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,69 +8,294 @@
 
 Simple middleware-style router
 
-## Installation
-
-```sh
+```bash
 $ npm install router
 ```
 
 ## API
 
 ```js
-var Router = require('router')
-```
-
-This module is currently an extracted version from the Express 4.x project,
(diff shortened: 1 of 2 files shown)


dougwilson:
I'm trying to read back over this, what was the plan here? I see at least that there was waiting to add tests to this PR, but no tests are here yet. The merge button is disabled due to conflicts as well. If you like, please get a test added / conflict resolved and open up a new PR. We're building out a 2.0 release plan now.

PR #13, opened 2015-01-21
Expose router engine for other implementations

### LINKED ISSUE #12: Router Engine
I need to implement a router similar to this one, but with different path matching semantics. Is there any plan to abstract the core logic into an engine module that other implementations can use?

blakeembrey:
There's still a couple of things to decide, but it works wonderfully for what I need and can plug any other implementation in that supports the same function return syntax. Only issue I see is debugging since it currently relies heavily on human-readable paths to be passed to the engine.

Closes #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,11 @@ $ npm install router
 
 ```js
 var finalhandler = require('finalhandler')
-var http         = require('http')
-var Router       = require('router')
+var http = require('http')
+var Router = require('router')
 
 var router = Router()
+
 router.get('/', function (req, res) {
   res.setHeader('Content-Type', 'text/plain; charset=utf-8')
   res.end('Hello World!')
@@ -179,11 +180,11 @@ router.route('/')
 
 ```js
 // import our modules
-var http         = require('http')
-var Router       = require('router')
+var http = require('http')
+var Router = require('router')
 var finalhandler = require('finalhandler')
-var compression  = require('compression')
-var bodyParser   = require('body-parser')
+var compression = require('compression')
+var bodyParser = require('body-parser')
 
 // store our message to display
 var message = "Hello World!"
@@ -244,8 +245,8 @@ curl http://127.0.0.1:8080/api/set-message -X PATCH -H "Content-Type: applicatio
 ### Example using mergeParams
 
 ```js
-var http         = require('http')
-var Router       = require('router')
+var http = require('http')
+var Router = require('router')
 var finalhandler = require('finalhandler')
 
 // this example is about the mergeParams option
@@ -294,6 +295,79 @@ curl http://127.0.0.1:8080
 curl http://127.0.0.1:8080/such_path
 > such_path
 ```
+
+## Implementing Your Own Router
+
+Implementing a custom path matching library on top of this module is as easy as using `Router.Engine`. For example, to implement an "exact" path matching module, we can do this:
+
+```js
+var Engine = require('router').Engine
+var slice = Array.prototype.slice
+
+/**
+ * Accepts the path and some options we defined in our engine.
+ */
+function toFunction (route, options) {
+  if (!options.end) {
+    return function (path) {
+      var matches = path.substr(0, route.length) === route
+
+      return matches ? { path: path } : false
+    }
+  }
+
+  return function (path) {
+    return path === route ? { path: path } : false
+  }
+}
+
+/**
+ * The constructor must return the engine instance.
+ */
+function ExactRouter (options) {
+  return Engine.call(this, options)
+}
+
+/**
+ * Inherits from the engine prototype.
+ */
+ExactRouter.prototype = Object.create(Engine.prototype)
+
+/**
+ * Set up `Router#use` with our custom path matching implementation.
+ */
+ExactRouter.prototype.use = function () {
+  // Use a simple utility for supporting a single path argument like `router`.
+  var opts = Engine.sanitizeUse.apply(null, arguments)
+  var match = toFunction(opts.path, { end: false })
+
+  return Engine.prototype.use.call(this, opts.path, match, opts.callbacks)
+}
+
+/**
+ * Set up `Router#route` with our custom path patching implementation.
+ */
+ExactRouter.prototype.route = function (path) {
+  var match = toFunction(path, { end: true })
+
+  return Engine.prototype.route.call(this, path, match)
+}
+
+/**
+ * Set up all the router method shorthands.
+ */
+Engine.methods.forEach(function (method) {
+  ExactRouter.prototype[method] = function (path) {
+    var route = this.route(path)
+    route[method].apply(route, slice.call(arguments, 1))
+    return this
+  }
+})
+```
+
+Both the path matching function and the path itself must be passed into the `route` and `use` engine methods. This is for debugging, so `path` should be a human-readable path name. `Engine#use` also accepts an array of handlers to immediately include. The match function must return an object of `{ path: string, params: object }` or `false` if it didn't match.
+
+Note: The path matching utility should not throw errors. Decoding of parameters is handled by the engine.
 
 ## License
 
--- a/engine.js
+++ b/engine.js
@@ -1,0 +1,798 @@
+/*!
+ * router
+ * Copyright(c) 2013 Roman Shtylman
+ * Copyright(c) 2014 Douglas Christopher Wilson
+ * MIT Licensed
+ */
+
+'use strict'
+
+/**
+ * Module dependencies.
+ * @private
+ */
+
+var debug = require('debug')('router')
+var flatten = require('array-flatten')
+var methods = require('methods')
+var mixin = require('utils-merge')
+var parseUrl = require('parseurl')
+var setPrototypeOf = require('setprototypeof')
+var Layer = require('./lib/layer')
+var Route = require('./lib/route')
+
+/**
+ * Module variables.
+ * @private
+ */
+
+var slice = Array.prototype.slice
+
+/* istanbul ignore next */
+var defer = typeof setImmediate === 'function'
+  ? setImmediate
+  : function(fn){ process.nextTick(fn.bind.apply(fn, arguments)) }
+
+/**
+ * Expose `Router`.
+ */
+
+module.exports = Engine
+module.exports.Route = Route
+module.exports.sanitizeUse = sanitizeUse
+module.exports.methods = methods.concat('all')
+
+/**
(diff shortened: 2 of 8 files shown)


blakeembrey:
@dougwilson Thoughts on an implementation like this?

dougwilson:
Hey, sorry, I've just been busy and haven't been able to look over this changeset yet (it's really big). The only thing I've noticed so far is you deleted a few tests, signaling it's not backwards-compatible, but I haven't look at anything else just yet.

blakeembrey:
@dougwilson It should be backward compatible, those tests were testing implementation details by using `route.path` but that's the same note as the debugging issue I mentioned. Since we're using a function now, testing for `route.path` doesn't make so much sense and neither does logging it. So maybe we add a property to functions that represent `route.path` that can be logged?

dougwilson:
Gotcha. So after the change, what does `req.route` become? It people using it now cannot drop in this change set, then that's by definition not backwards-compatible :)

dougwilson:
For example, there seem to be many, many people using `req.route.path`, but I guess that's missing now? Since it was a publicly-exposed thing, if it's missing and all they did was upgrade this module, it would have to be under a major version bump is all. I'm not saying we cannot, I'm just saying what I saw _so far_ :)

blakeembrey:
The only change is that `req.route.path` no longer exists. If that's an issue with breaking backward compatibility then we could keep passing the path through easily. It'd fix the debugging issues too, to keep passing that around.

blakeembrey:
Ok, good to know. I'll add that functionality back now and ping you again :+1:

blakeembrey:
By the way, one thing I noticed was `strict: false` on `.use` instead of using the `this.strict` option - is that intentional?

blakeembrey:
@dougwilson No backward incompatibility now.

blakeembrey:
@dougwilson Sorry to be so pushy, but I really want to publish a bunch of middleware on top of this functionality this week. Do you have any thoughts on the implementation?

dougwilson:
Sure. Sorry, it's hard to look through " 801 additions and 739 deletions" :) Do you have any documentation of how this new feature would be used? (so I can at least maybe only review the external API :P).

dougwilson:
I also don't completely understand what you're doing with this `truth()` function in a few of the files.

blakeembrey:
The implementation layer is pretty much `index.js` and most of the changes are just moving `index.js` to `engine.js` and making the layers use a match function instead of regexp. It's not actually a lot of changes, but the files moving made it a bit messy (sorry!)

I can also publish the module implementation I've implemented to show you how I'm using it.

blakeembrey:
> I also don't completely understand what you're doing with this truth() function in a few of the files.

Yeah, sorry. Maybe there's a better way to document it, but it's basically the "fast match" route from before that you always expected to match (E.g. `use`). It's a function that always evaluates to true, but I'll think if there's a better implementation. Thoughts?

dougwilson:
> I can also publish the module implementation I've implemented to show you how I'm using it.

That's fine, and we can also exchange it privately if you feel that necessary :) But we'll probably want docs in the readme eventually, haha.

blakeembrey:
I can try my hand at some simple documentation at the bottom of the readme.

dougwilson:
> but I'll think if there's a better implementation. Thoughts?

I don't know enough about the implementation to really comment :)

dougwilson:
From the title:

> Expose router engine for other implementations 

Based on your file names, I would expect that you'd have an exports of `Engine`? I don't see that, though?

blakeembrey:
> Based on your file names, I would expect that you'd have an exports of Engine? I don't see that, though?

Like in `index.js`? I just expected people to `require('router/engine')`, but you wouldn't?

dougwilson:
> I just expected people to require('router/engine'), but you wouldn't?

No, I wouldn't; I don't consider that to be a public API, otheriwse I'd never be free to rearrange any files in my package.

blakeembrey:
> No, I wouldn't; I don't consider that to be a public API, otheriwse I'd never be free to rearrange any files in my package.

That makes sense, maybe there needs to be a `router-engine` module then - which this implements?

dougwilson:
Perhaps, but please don't take it :) I was just wondering. I'm going to be up for a long time tonight, so I'm going to take a serous look at this, because I share the same want as you :)

dougwilson:
Bu yea, don't mind the exports, then :) If it definitely is agreeable that this would be a `route/engine` case, I can easily split it on merge :) I just wanted to make sure you intended to have people use what's in the `./engine` file :)

blakeembrey:
> Perhaps, but please don't take it :) I was just wondering. I'm going to be up for a long time tonight, so I'm going to take a serous look at this, because I share the same want as you :)

Ok, awesome! Thanks so much :smile: I'm looking forward to releasing new modules on top of this functionality :+1:

blakeembrey:
@dougwilson I pushed the module I built that shows how other implementations would use the engine: https://github.com/mulesoft-labs/osprey-router/blob/master/osprey-router.js

blakeembrey:
@dougwilson I ended up building everything on top of this (https://github.com/mulesoft/osprey) in a bet it (or something similar) will eventually exist. The only other thing I need is to be able to skip the rest of the stack which I saw was already an issue in `express`.

dougwilson:
Oh, cool. So yes, there is an open issue in Express, and it's mainly because of the naming of the special string (`router`); I had decided "whatever, let's just do it even if they sounds similar" but then realized that reading "`next('route')`" generally makes sense, "`next('router')`" doesn't really (it's more of a "`exit('router')`" or something). If you can weigh in on the naming, I can get that in, at least, because the implementation itself is simple and straight-forward.

blakeembrey:
Ok, noted. Can't think of anything better than `end/finish/exit` right now.

I'll probably ping you in a few days about the engine, not sure if you've managed to take a look yet :stuck_out_tongue_closed_eyes:

blakeembrey:
@dougwilson Any luck yet?

dougwilson:
I definitely haven't forgotten :) At least recently, getting everything tested for Node.js 0.12 ended up taking priority :D

blakeembrey:
@dougwilson Ping (sorry)

blakeembrey:
@dougwilson Hope it's ok to bother you once again :smile:

dougwilson:
It's always OK and I'm sorry for the delays, of course :(

blakeembrey:
Any chance this will happen or should I look at branching it and doing a new PR myself? @dougwilson

dougwilson:
It'll definitely happen, haha, I'm sorry :( I was doing some things tonight, but without real direction. Let me look into this tonight (at least the contents of this PR) and if nothing substantial comes out of tonight, then I'll just default to accepting whatever dependency you wanted to stand up (or I guess even this refactoring) :)

blakeembrey:
@dougwilson Thanks for the quick response. Didn't quite grok the last sentence, but I'll need to refactor this PR since it won't merge right now. Should we create a separate "engine" module?

dougwilson:
Oh, sorry. So it meant that if I don't have anything done tonight or any valuable feedback, then I'll just end up accepting whatever PR you provide (if that's a PR that just refactors the engine into a file like this one, or as another dependency, then so be it). If it does end up being a dependency, ideally it's in this org, which you are an owner in :)

blakeembrey:
@dougwilson Just rebased, wasn't at painful (luckily). Only reason there was a conflict is because 99% of the logic from `router.js` is now in `engine.js`. I've kept it here for now since the tests are super comprehensive and to do that amount of tests again, I imagine, would end up implementing this router as the example implementation anyway.

blakeembrey:
@dougwilson Ok, done. I copied some tests over, but the bulk of the changes I just made were because I realised there were some simple optimisations to do. Moved `fast_slash` and `decode_param` to be engine implementations, not router implementations and also realised that every router implementation doesn't need to define all the methods, just needs to override `Engine#route`.

blakeembrey:
Actually, I shouldn't be coding at 1am. The reason people needed to implement these specifically was because other implementations can use a different number of arguments. However, they only difference for those users now will be that they can't rely on the methods already set up and can't use the `sanitizeUse` utility I added. I guess the question now is, is it better to be explicit or implicit for other implementors? We can always add a simple note saying that if the implementors accepts a different number of arguments, they need to override the method implementations. Thoughts? Then this is done :smile:

dougwilson:
Nice! And yes, that is my main concern: that the changes fit into a case where people can use it to build upon :)

blakeembrey:
@dougwilson Well, if it helps, we've been using it in https://www.npmjs.com/package/osprey-router for the past few months. All the methods can be overridden if you need to support for arguments and I'll make that note in the README tomorrow. I'll remove the method implementations by default and add it back to the example, instead exposing the array from `methods` for implementors. Was there anything else you were thinking about?

blakeembrey:
@dougwilson Ok, all done and dusted. The sample implementation in the README should be adequate for anyone else to get started.

blakeembrey:
@dougwilson Have you had any more time to review this at all? Would love to re-open the discussion around unicode URLs once this is merged.

blakeembrey:
@dougwilson Any chance you could review this before too many more changes continue happening :stuck_out_tongue_closed_eyes:

blakeembrey:
@dougwilson :pray:

dougwilson:
Sorry, I have just been so busy herding all these modules around :) I just finished looking through a lot of the Express 4.x issues, so this is on my backlog after getting 4.13.0 out to get this up to snuff and then into the 5.x branch of Express very soon to get a 5.0 beta out! I consider this issue a blocker to 5.0 beta and we really need to get 5.0 out next month, so I hope that signals the importance/timeline of this issue :)

blakeembrey:
@dougwilson Just rebased back against the latest `master` release. Still looking forward to merge :stuck_out_tongue_winking_eye:

I just realised (maybe I did before) but this module still uses `path-to-regexp@0.1.x`. Shouldn't it be using `1.x` now? Or will that be updated before the `5.0` release with this?

blakeembrey:
@dougwilson Rebased against `1.1.3`. Anything I can do to speed this along? By the way, the best way to diff this is to compare the new `engine.js` against the old `index.js`. That should make diffing it easier, since things are pretty much the same minus a couple of parts to decouple the engine.

blakeembrey:
@dougwilson I need to continue working on this branch for promise support and router exiting, so I'm going to close this and reopen under a new PR.

Issue #14, opened 2015-02-02
Advertise Browser Support


wesleytodd:
I am interested to know if you all think it is worth mentioning that this module is compatible with browsers.  I threw together an example of using this module in front-end code, and it works perfectly fine.

It is a big deal to say that the module is "Browser Compatible" because of the support issues, but the current implementation works in Chrome & FF (which is all I tested).  And after reading most of the code I don't see anything that will clearly break it in other browsers (at least with the required polyfills).

dougwilson:
We could do it if someone is willing to help me setup automated browser testing (because otherwise it's just too easy to accidentally break browser support). I just have never setup browser testing before.

wesleytodd:
I believe we had something setup internally for this, probably with Sauce Labs, I will ask our QA lead if he has any suggestions and post back.

Really I am just glad to hear that you are interested in this being a "supported feature" of the module.  I am going to write a push-state router implementation wrapping this module, and didn't want to base my stuff on something that might not be supported in the long run.

dougwilson:
No problem. There are modules that are definitively Node.js-specific, and others (like this) that are basically neutral (that don't intend to not work in browsers, but it's not the author's primary concern). If I can set it up so I get automated browser tests, then I'll know when a change breaks and can just make it not break, thus providing support :)

rlidwka:
I'm curious, what it can be used for in browser environment? Some kind of a single-page app?

wesleytodd:
@rlidwka Yeah, we are doing an isomorphic app.  Basically I am making a `routes.js` file that looks like this:

```javascript
var index = require('./handlers/index'),
    foo = require('./handlers/foo');

module.exports = function(router) {
    router.get('/', index);
    router.get('/foo/:bar', foo);
};
```

Where `router` is an express app on the server and a wrapper around this router on the client.  Then I make each middleware a module and use browserify's `browser` field in the package.json to load the client or server side version of the middleware.

We are using React for rendering and a wrapper around XHR/Request or data loading.  It enables almost all of our modules to work on both the server and the client.

wesleytodd:
For backstory:

Our last application was an Angular app with a Go backend.  One of the most common bugs we had was weird behavior because we forgot to define all the proper routes in both Go and javascript.  This was both a consequence of our architecture (multiple angular apps on a single domain) and having the routes duplicated.  So this time around we really want to avoid the duplicated route issue.

wesleytodd:
[...]

PR #17, opened 2015-04-22
Fixed setPrototypeOf for middle aged browsers

wesleytodd:
As discussed in #14

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "methods": "~1.1.1",
     "parseurl": "~1.3.0",
     "path-to-regexp": "0.1.3",
+    "setprototypeof": "~1.0.0",
     "utils-merge": "1.0.0"
   },
   "devDependencies": {
--- a/index.js
+++ b/index.js
@@ -17,6 +17,7 @@ var methods = require('methods')
 var mixin = require('utils-merge')
 var parseUrl = require('parseurl')
 var Route = require('./lib/route')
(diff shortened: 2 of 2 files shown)


wesleytodd:
I manually tested this in OSx (chrome, ff, safari) and Windows (chrome, ff, IE10-11).  Works just fine as long as you have the right poly fills.  Here are the ones I included: https://github.com/wesleytodd/nighthawk/blob/master/lib/polyfills.js

I didn't test IE9 because it doesn't support the History api and has based routing is not something that I think should exist anymore.  But if someone has something that wants to support that they can go ahead and test it.

package.json
```diff
@@ -11,6 +11,7 @@
     "methods": "~1.1.1",
     "parseurl": "~1.3.0",
     "path-to-regexp": "0.1.3",
+    "setprototypeof": "^1.0.0",
```
dougwilson: remove the `^` for old npm versions :)

wesleytodd:
Fixed!!

dougwilson:
Nice, looks like your module is exactly this: https://github.com/pillarjs/router/issues/14#issuecomment-72785898

I approve and will merge + release :)

wesleytodd:
Yep, that is exactly what I modeled after.  It was the best and simplest I could find.

dougwilson:
Published as 1.1.0

wesleytodd:
Awesome, thanks!!

PR #23, opened 2015-09-28
Refactoring for ES6 harmony functional

nooks:


dougwilson:
Hi! This module has no intention of dropping Node.js 0.10 support any time soon. Would these changes even work on Node.js 0.10? The tests seem to fail on every single supported version of Node.js for this module.

nooks:
I suggest to open a branch 2.0 for Node.js Next, which will normally be maintained spread operators. I'll try for the sake of experimentation to invest your contribution)

nooks:
Example for ***sendOptionsResponse***
ECMAScript 6 (Use spread and Set)
```javascript
let sendOptionsResponse = (res, methods) => {
  let options = new Set(methods)

  // construct the allow list
  let allow = [...options].sort().join(', ')

  // send response
  res.setHeader('Allow', allow)
  res.setHeader('Content-Length', Buffer.byteLength(allow))
  res.setHeader('Content-Type', 'text/plain')
  res.setHeader('X-Content-Type-Options', 'nosniff')
  res.end(allow)
}
```
ECMAScript 5
```javascript
function sendOptionsResponse(res, methods) {
  var options = Object.create(null)

  // build unique method map
  for (var i = 0; i < methods.length; i++) {
    options[methods[i]] = true
  }

  // construct the allow list
  var allow = Object.keys(options).sort().join(', ')

  // send response
  res.setHeader('Allow', allow)
  res.setHeader('Content-Length', Buffer.byteLength(allow))
  res.setHeader('Content-Type', 'text/plain')
  res.setHeader('X-Content-Type-Options', 'nosniff')
  res.end(allow)
}
```
Example for ***wrap***
ECMAScript 6 (Use spread)
```javascript
let wrap = (old, fn) => {
  return (...args) => {
    fn(old, ...args)
  }
}
```
ECMAScript 5
```javascript
function wrap(old, fn) {
  return function proxy() {
    var args = new Array(arguments.length + 1)

    args[0] = old
    for (var i = 0, len = arguments.length; i < len; i++) {
      args[i + 1] = arguments[i]
    }

    fn.apply(this, args)
  }
}
```

_If there were benchmarks, I'd measured the performance_

dougwilson:
The next release, 2.0, is still going to be Node.js 0.10. That would mean for this, I would have to maintain three releases of this module at once, which I don't want to do. Can you tell me why I would want to use ES6 anyway? What are the benefits?

nooks:
1) Fast arrow functions
2) Map and Set faster than the Object itself and Array
3) spread operator faster than the arguments

nooks:
Why 0.10 instead of 0.12?

nooks:
Express 5.0 базируется на node.js 0.12 и будет использовать router?

dougwilson:
Can you show benchmarks comparing this module as published to npm vs your PR?

> Why 0.10 instead of 0.12?

Most corporate PaaS systems only just upgraded to Node.js 0.10 and 0.10 is still being supported for at least another year. The organization pillarjs is here to house the building blocks of frameworks. Because of this, all modules will always support all active (even ones only in maintenance mode) branches of Node.js.

I'm happy to keep this PR open the 1 year+ until Node.js 0.10 is EOL and reconsider then.

Having a higher minimum Node.js version belongs to the high-level frameworks, like Express, not to the low-level pieces like this module.

> Express 5.0 is based on node.js 0.12 and will use the router?

Express 5 will be based on Node.js 0.10 or 4, depending on discussions. That does not have any impact on this module, which is used by other frameworks. The Node.js support policy for this module is all active (even ones only in maintenance mode) branches of Node.js.

There is also a bias here in that I, as maintainer of these modules, require them to be usable within my own corporate environment which is Node.js 0.10 for the foreseeable future (and I don't have control over that).

nooks:
Close PR, benchmarks are not paid off. Chances are yet to optimize the arrow functions, they need to insert inline function, as it is done with the usual functions without context. Since this uses the arrow functions and arguments out of context where its cause

nooks:
I'm sorry to trouble you

PR #24, opened 2015-10-01
Minor refactors

calebmer:
Two minor refactors to make the code more consistent with the express code base.

- Move the `Router` class definition into the `lib` folder. `index.js` now simply requires that file (the way it was previously is confusing for new developers)
- Refactored snake_case names to camelCase


dougwilson:
The rule for code here is that the bulk of all code must be in the `index.js` file, if you can undo that change, please.

The snake_case exists because of backwards-compatibility, so altering those means that there is no longer any promised release timeline for the changes, since they are not backwards-compatible.

dougwilson:
Typically, if you do intend to refactor code around in other projects, it's fine, but you should go ahead and open an issue listing exactly what you are intended to refactor, the reason for each refactor, and get some buy-in first so you do not get surprise answers after doing a bunch of work, if that makes sense.

calebmer:
Ok, thanks for being understanding. Typically I don't work with projects with lots of legacy code so I did not think about these things.

ChiperSoft:
Legacy code! Router was written for Express 5, its still new!

PR #29, opened 2015-11-12
Expose router engine for other implementations

blakeembrey:
From https://github.com/pillarjs/router/pull/13.

Closes https://github.com/pillarjs/router/pull/12.

Implements custom path implementations via router inheritance.

Edit: I also still prefer the simplicity of this PR over https://github.com/pillarjs/router/pull/22. If you think about it this way it's cleaner - the router engine is the most basic router which takes a function as the "path matcher". Other implementations can build on this to take strings and regexps, etc. The path accepted for `route` and `use` can be though of as the "label" for debugging (and backward compat).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,11 @@ $ npm install router
 
 ```js
 var finalhandler = require('finalhandler')
-var http         = require('http')
-var Router       = require('router')
+var http = require('http')
+var Router = require('router')
 
 var router = Router()
+
 router.get('/', function (req, res) {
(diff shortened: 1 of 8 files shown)


blakeembrey:
@dougwilson Any chance we'll be able to review these sometime? It's been a more than a year (https://github.com/pillarjs/router/pull/13) :open_mouth: Otherwise I can look at pulling out the inheritance pieces, I'd love to get a more stable base to continue using (https://www.npmjs.com/package/osprey).

PR #33, opened 2015-12-30
Update .travis.yml

olegabr:
test for latest node versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ node_js:
   - "4.0"
   - "4.2"
   - "5.0"
+  - "5.1"
+  - "5.2"
+  - "5.3"
 sudo: false
 before_install:
   # Setup Node.js version-specific dependencies


dougwilson:
Hi! It looks like Node.js 5.1+ all fail their tests. If you would like us to bump the version, please update the PR to fix the test failures in those versions :)

olegabr:
Hi! It was just a quick test if I can use Router in my nodejs 5.3 project. The answer is No for now.
I have no skill to solve the problem, but there is only one broken test: `Router should reject missing callback`:
```js
  it('should reject missing callback', function () {
    var router = new Router()
    assert.throws(router.bind(router, {}, {}), /argument callback is required/)
  })
```

```
  1 failing
  1) Router should reject missing callback:
     SyntaxError: Invalid regular expression: /^function router(req,res,next){__cov_QHUO0qwDsAlzMrhf2dkKMw\.f['3']++;__cov_QHUO0qwDsAlzMrhf2dkKMw\.s['19']++;router\.handle(req,res,next);}\/?$/: Nothing to repeat
      at new RegExp (native)
      at pathtoRegexp (node_modules/path-to-regexp/index.js:128:10)
      at new Layer (lib/layer.js:9:1194)
      at Function.route (index.js:9:19032)
      at Function.Router.(anonymous function) [as bind] (index.js:9:19729)
      at Context.<anonymous> (test/router.js:25:26)
```

It is non-clear for me, but may be some of maintainers understend the reasons...

dougwilson:
Looks like the source is that in 5.1, Node.js added support for the BIND method. Since we declare all methods as methods on the router instance, we end up shadowing the JavaScript native bind method.

dougwilson:
Version 1.1.4 of this module (publishing soon) tests fine against Node.js 5.4 now.

PR #34, opened 2016-02-13
Store matched routes in request

ajfranzoia:
I'm transplanting this PR from the original I made at https://github.com/expressjs/express/pull/2864 (issue discussed at https://github.com/expressjs/express/issues/2501).

It allows to store every matched route in a request when travelling through nested routes. This may be useful for logging, ACL checks, etc.
So, for example, if you have your routing set like this:

``` javascript
var fooRouter = new Router();
var barRouter = new Router();
var bazRouter = new Router();

bazRouter.get(['/bez', '/baz/:subId'], function(req, res, next) {
  next();
});

fooRouter.use(['/foo/:id', '/foe'], barRouter);
barRouter.use(['/bar'], bazRouter);
```

A request for the URL `/foo/10/bar/baz/30` will set an array of matched routes in the request object like this:

``` javascript
req.matchedRoutes = ['/foo/:id', '/bar', '/baz/:subId'];
```

Let me know if there is anything that needs to be modification, unmeaningful variable names, etc.

### DIFF
--- a/index.js
+++ b/index.js
@@ -286,6 +286,11 @@ Router.prototype.handle = function handle(req, res, callback) {
         return next(layerError || err)
       }
 
+      // store matched routes
+      if (layer.path) {
+        req.matchedRoutes = (req.matchedRoutes || []).concat(layer.matchedPath.path)
+      }
+
       if (route) {
         return layer.handle_request(req, res, next)
       }
@@ -341,7 +346,7 @@ Router.prototype.process_params = function process_params(layer, called, req, re
   var params = this.params
 
   // captured parameters from the layer, keys and values
-  var keys = layer.keys
+  var keys = layer.matchedPath ? layer.matchedPath.keys : []
 
   // fast track
   if (!keys || keys.length === 0) {
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -28,23 +28,32 @@ var hasOwnProperty = Object.prototype.hasOwnProperty
 
 module.exports = Layer
 
-function Layer(path, options, fn) {
+function Layer(paths, options, fn) {
   if (!(this instanceof Layer)) {
-    return new Layer(path, options, fn)
+    return new Layer(paths, options, fn)
   }
 
-  debug('new %s', path)
+  debug('new %s', paths)
   var opts = options || {}
 
   this.handle = fn
   this.name = fn.name || '<anonymous>'
   this.params = undefined
   this.path = undefined
-  this.regexp = pathRegexp(path, this.keys = [], opts)
 
-  if (path === '/' && opts.end === false) {
-    this.regexp.fast_slash = true
+  if (paths === '/' && opts.end === false) {
+    this.fastSlash = true
   }
+
+  this.paths = !Array.isArray(paths) ? [paths] : paths
+  this.paths = this.paths.map(function (path) {
+    var pathObj = {
+      path: path,
+      keys: []
+    }
+    pathObj.regexp = pathRegexp(path, pathObj.keys, opts)
+    return pathObj
+  })
 }
 
 /**
@@ -113,14 +122,23 @@ Layer.prototype.match = function match(path) {
     return false
   }
 
-  if (this.regexp.fast_slash) {
+  if (this.fastSlash) {
     // fast path non-ending match for / (everything matches)
     this.params = {}
     this.path = ''
+    this.matchedPath = this.paths[0]
     return true
   }
 
-  var m = this.regexp.exec(path)
+  var checkPath, m;
+
+  for (var i = 0; i < this.paths.length; i++) {
+    checkPath = this.paths[i]
+    if (m = checkPath.regexp.exec(path)) {
+      this.matchedPath = checkPath
+      break
+    }
+  }
 
   if (!m) {
     this.params = undefined
@@ -137,7 +155,7 @@ Layer.prototype.match = function match(path) {
   var params = this.params
 
   for (var i = 1; i < m.length; i++) {
-    var key = keys[i - 1]
+    var key = this.matchedPath.keys[i - 1]
     var prop = key.name
     var val = decode_param(m[i])
 
--- a/test/router.js
+++ b/test/router.js
@@ -985,6 +985,31 @@ describe('Router', function () {
   })
 })
 
+describe('req.matchedRoutes', function() {
+  it('should store matchedRoutes in request', function(done) {
+    var router = new Router()
+    var barRouter = new Router()
+    var bazRouter = new Router()
+    var server = createServer(router)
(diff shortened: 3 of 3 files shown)


index.js
```diff
@@ -286,6 +286,11 @@ Router.prototype.handle = function handle(req, res, callback) {
         return next(layerError || err)
       }
 
+      // store matched routes
+      if (layer.path) {
+        req.matchedRoutes = (req.matchedRoutes || []).concat(layer.matchedPath.path)
```
wesleytodd: This is going to create unnecessary arrays.  Can you instead only create a new array the first time, then push to it after?

```javascript
req. matchedRoutes = req.matchedRoutes || []
req.matchedRoutes.push(layer.matchedPath.path)
```

lib/layer.js
```diff
@@ -28,23 +28,32 @@ var hasOwnProperty = Object.prototype.hasOwnProperty
...
   this.handle = fn
   this.name = fn.name || '<anonymous>'
   this.params = undefined
   this.path = undefined
-  this.regexp = pathRegexp(path, this.keys = [], opts)
 
-  if (path === '/' && opts.end === false) {
-    this.regexp.fast_slash = true
+  if (paths === '/' && opts.end === false) {
+    this.fastSlash = true
```
wesleytodd: V8 can do optimizations if the shape of an object remains the same.  This addition will have only some `Layer`'s with a `fastSlash` property.  To optimize this, you should define it as `undefined` above, or `false` here.

lib/layer.js
```diff
@@ -28,23 +28,32 @@ var hasOwnProperty = Object.prototype.hasOwnProperty
...
   }
 
-  debug('new %s', path)
+  debug('new %s', paths)
   var opts = options || {}
 
   this.handle = fn
   this.name = fn.name || '<anonymous>'
   this.params = undefined
   this.path = undefined
```
wesleytodd: Add a line for `this.matchedPath = undefined`, see comment below for reasoning.

lib/layer.js
```diff
@@ -113,14 +122,23 @@ Layer.prototype.match = function match(path) {
...
+  if (this.fastSlash) {
     // fast path non-ending match for / (everything matches)
     this.params = {}
     this.path = ''
+    this.matchedPath = this.paths[0]
     return true
   }
 
-  var m = this.regexp.exec(path)
+  var checkPath, m;
```
wesleytodd: All of the express modules are working toward compliance with `standard`.  Which includes the rule that variable declarations are each on a single line.  We have doing it piecemeal because it would cause too many conflicts otherwise, so since this is a new line, can you make the change just here?

test/router.js
```diff
@@ -985,6 +985,31 @@ describe('Router', function () {
   })
 })
 
+describe('req.matchedRoutes', function() {
```
wesleytodd: Same as comment above, can you make the formatting through here `function () {`?  As per `standard`.

wesleytodd:
Also, maybe we should change the property name to `matchedPaths`.  There is something called a `Route`, and it may be less ambiguous to call it this.  Thoughts on that?

GuskiS:
Is there a plan to get this merged?

EgilsZ:
Can someone please give any timeline on when this could be merged? Anything blocking this to be completed?

wesleytodd:
Nothing blocking AFAIK, there are just higher priorities.  If you are interested in helping out you can take a look at these and start participating in the project:

https://github.com/expressjs/express/blob/master/Contributing.md
https://github.com/expressjs/express/pull/4055

More people helping collaborate and traige issues would unblock the core team and allow us to spend more time on moving these features forward.  Thanks in advance for your help!

saasen:
We would love this feature, as we're trying to expose Prometheus metrics based on the **resolved** route, rather than the incoming URL and normalizing it (which is very error prone).

wesleytodd:
@saasen, my original interest in this feature was the exact same as your use case.  If you are looking for a good solution which is simple and does not have many edge cases, you can do what I did in that app:

```javascript
const onFinished = require('on-finished')
function routeMetrics (route) {
  return [route, (req, res) => {
    onFinished(res, (err) => {
      reportMetric(err, route)
    })
  }]
}

app.get(routeMetrics('/:foo'), routeHandlerMiddleware)
```

The only thing to be aware of with a solution like this is that if a request would match two routes and both get run, both would report.  You can prevent this if it is a concern for your app by setting a flag on `res` like `res.locals.route = route` and then reference that off of `res` inside the `onFinished` handler function.  Make sense?

saasen:
@wesleytodd: Would this work when defining routes inside a Router as well? AFAIK, the route resolved inside a Router doesn't contain the "parent"-route.

dougwilson:
For the folks who keep asking when it will be merged / what is blocking: there are open review issues at the top that haven't been addressed yet.

wesleytodd:
@saasen Hm, good question.  Without actually trying that I am guessing no and that you would need to use a whole express app which has a `mountpath`.  Maybe there is another way I am missing right now.

@dougwilson Oh yeah, I guess I did open a review!  I wonder if @ajfranzoia is interested in making those updates?  If not I might be able to find some time soon to take this over and clean it up.

gabegorelick:
I rebased these changes into #86. But it's mostly unchanged from this PR.

index.js
gabegorelick: Fixed in https://github.com/pillarjs/router/pull/86/files#diff-168726dbe96b3ce427e7fedce31bb0bcR295

lib/layer.js
gabegorelick: Fixed in https://github.com/pillarjs/router/pull/86/files#diff-be883fbed8968009576c828e0ff70e36R43

lib/layer.js
gabegorelick: Fixed in https://github.com/pillarjs/router/pull/86/files#diff-be883fbed8968009576c828e0ff70e36R47

lib/layer.js
gabegorelick: Fixed in https://github.com/pillarjs/router/pull/86/files#diff-be883fbed8968009576c828e0ff70e36R121

test/router.js
gabegorelick: Should be fixed in https://github.com/pillarjs/router/pull/86/files#diff-c6bcc5b8d170df93e5b3223fa590f1f7R1133

gabegorelick:
Sorry for the noise, but wanted to mark off all of the feedback from @wesleytodd to ensure that it's incorporated into #86.

dougwilson:
Closing this in favor of the progress being made in https://github.com/pillarjs/router/pull/86

PR #37, opened 2016-02-24
Make req.route an object saving route and baseUrl

tunniclm:
This PR attempts to implement https://github.com/expressjs/express/issues/2883.

----

When using an error handler you have access to `req.route`
which describes the most recently matched route. However,
some other informaiton (like `req.baseUrl`) gets changed
before the error handler runs.

This commit changes `req.route` to an object that also contains
the `req.baseUrl` before it gets changed.

Since the actual `route` object is not scoped to the current
event, this commit instead uses a new object to store the
`baseUrl` but sets the `__proto__` to the `route` object so that
all its properties are also available.

### DIFF
--- a/index.js
+++ b/index.js
@@ -271,7 +271,8 @@ Router.prototype.handle = function handle(req, res, callback) {
 
     // store route for dispatch on change
     if (route) {
-      req.route = route
+      req.route = req.route || { __proto__: route }
+      req.route.baseUrl = req.baseUrl;
     }
 
     // Capture one-time layer values
--- a/lib/route.js
+++ b/lib/route.js
@@ -104,7 +104,8 @@ Route.prototype.dispatch = function dispatch(req, res, done) {
     method = 'get'
   }
 
-  req.route = this
+  req.route = req.route || { __proto__: this }
+  req.route.baseUrl = req.baseUrl;
 
   next()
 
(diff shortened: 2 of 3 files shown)


tunniclm:
Some thoughts/questions:
- Is using `__proto__` ok? It could instead copy the values to a new object `{}`, or alternatively should it be a new object with a differnet name, such as `req.routeInfo`, that contains properties `req.routeInfo.route` and `req.routeInfo.baseUrl`?
- Is the test okay? Should it instead (or additonally) test the value of `req.route.baseUrl` in an error handler? Should it use a different style of test where the value checked is in the body of the response rather than a header?

index.js
```diff
@@ -271,7 +271,8 @@ Router.prototype.handle = function handle(req, res, callback) {
 
     // store route for dispatch on change
     if (route) {
-      req.route = route
+      req.route = req.route || { __proto__: route }
+      req.route.baseUrl = req.baseUrl;
```
tunniclm: Oops, must remove the habitual `;`

tunniclm:
Any thoughts/feedback on this?

wesleytodd:
Hey! I am (finally) getting around to trying to wrangle these and I think that this is a good idea. Sadly I think it is breaking and since it is so old I am not sure it is a good idea to revive this as is or take a new try specifically on top of the `2.0` branch? If you are interested in landing this for our push toward `express@5` I would love to chat. If not, I think it will need to wait for v6. Let me know asap so I can work it into the plans.

Issue #41, opened 2016-05-19
router.bind is overridden as an http method


wesleytodd:
Not sure if this is a new things in node 6, I'm running `6.1.0`, but the core `http.methods` include the http `BIND` method.  When we loop the methods we end up overriding `function.bind` on the router instance causing weird behavior.  The test that broke was as follows:

```javascript
it('should reject missing callback', function () {
  var router = new Router()
  assert.throws(router.bind(router, {}, {}), /argument callback is required/)
})
```

The intent here is obvious, but it ended up calling the [method setup here](https://github.com/pillarjs/router/blob/master/index.js#L529-L535).

There are three solutions I can see:

1. Filter out known bad names (aka blacklist)
2. Only generate known good methods (aka whitelist)
3. Document the fact that we overrode `.bind`

I can make a PR either way but wanted to open the discussion first.

Also for reference, here is the list I got by logging the methods array:

```
[ 'acl',
  'bind',
  'checkout',
  'connect',
  'copy',
  'delete',
  'get',
  'head',
  'link',
  'lock',
  'm-search',
  'merge',
  'mkactivity',
  'mkcalendar',
  'mkcol',
  'move',
  'notify',
  'options',
  'patch',
  'post',
  'propfind',
  'proppatch',
  'purge',
  'put',
  'rebind',
  'report',
  'search',
  'subscribe',
  'trace',
  'unbind',
  'unlink',
  'unlock',
  'unsubscribe' ]
```

dougwilson:
So since I noticed this from a 5.x release back in January, I've through a bunch about it, because it affects this module and Express itself. I think that if we want to continue to do `router[method]` style, the methods should be a static list, not dynamic, and we should have an additional API that takes a method name so people can use any method they want to use as well. This would also have the benefit for things like TypeScript and JSDoc comments to actually function with the methods, since they won't be auto-generated.

wesleytodd:
Ok, so maybe we should bet a 2.0 roadmap that includes this and #42 at the minimum.  I can work on this over the next few days and put together the PR.  

I will open a new issue to discuss supporting something where the method is passed in so we can focus this issue on specifying the static method list.

PR #42, opened 2016-05-19
Updated to path-to-regexp@3.0.0

wesleytodd:
This would probably be a major version bump for router since the path matching semantics are so different.

Also. you will notice a change in the tests.  I opened a separate issue to discuss, #41.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,64 @@ router.route('/pet/:id')
 server.listen(8080)
 ```
 
+## Migrating to 2.x from 1.x
+
+The main change is the update to `path-to-regexp@2.0.0`, which has a few breaking changes:
+
+#### No longer a direct conversion to a RegExp with sugar on top.
+
+It's a path matcher with named and unnamed matching groups.  It's unlikely you previously abused this feature,
+it's rare and you could always use a RegExp instead. An example of this would be:
+
+```javascript
+// Used to work
+router.get('/\\d+')
+
+// Now requires matching group
+router.get('/(\\d+)')
+```
+
+#### All matching RegExp special characters can be used in a matching group.
+
+Other RegExp features are not supported - no nested matching groups, non-capturing groups or look aheads
+There is really only one common change is needing replacing any routes with `*` to `(.*)`.  Some examples:
+
+- `/:user(*)` becomes `/:user(.*)`
+- `/:user/*` becomes `/:user/(.*)`
+- `/foo/*/bar` becomes `/foo/(.*)/bar`
+
+#### Parameters have suffixes that augment meaning - `*`, `+` and `?`. E.g. `/:user*`
+
+Needs more info.
+
+#### Named params with regex no longer define positionally.
+
+One other small change (hopefully low impact), is that named parameters with regular expressions no longer result in positional
+values in the `params` object.  An example is:
+
+```javascript
+router.get('/:foo(.*)')
+
+// old GET /bar
+console.log(req.params) // {0: 'bar', 'foo': 'bar'}
+
+// new GET /bar
+console.log(req.params) // {'foo': 'bar'}
+```
+
+#### Partial matching, prefer escaping delimiter
+
+The update to `path-to-regexp@3` includes a change to how partial matches are handled,
+now you should escape the delimiter before a partial match segment.  For example:
+
+```javascript
+// old
+router.get('/user(s)?/:user/:op')
+
+// new
+router.get('\\/user(s)?/:user/:op')
+```
+
 ## License
 
 [MIT](LICENSE)
--- a/test/route.js
+++ b/test/route.js
@@ -645,7 +645,7 @@ describe('Router', function () {
         it('should work following a partial capture group', function (done) {
           var cb = after(2, done)
           var router = new Router()
-          var route = router.route('/user(s)?/:user/:op')
+          var route = router.route('\\/user(s)?/:user/:op')
           var server = createServer(router)
 
           route.all(sendParams)
@@ -716,7 +716,7 @@ describe('Router', function () {
 
         it('should capture everything with pre- and post-fixes', function (done) {
           var router = new Router()
-          var route = router.route('/foo/*/bar')
+          var route = router.route('/foo/(.*)/bar')
           var server = createServer(router)
 
           route.all(sendParams)
@@ -728,7 +728,7 @@ describe('Router', function () {
 
         it('should capture greedly', function (done) {
           var router = new Router()
-          var route = router.route('/foo/*/bar')
+          var route = router.route('/foo/(.*)/bar')
           var server = createServer(router)
 
           route.all(sendParams)
@@ -740,7 +740,7 @@ describe('Router', function () {
 
         it('should be an optional capture', function (done) {
           var router = new Router()
-          var route = router.route('/foo*')
+          var route = router.route('/foo(.*)')
           var server = createServer(router)
 
           route.all(sendParams)
@@ -753,7 +753,7 @@ describe('Router', function () {
         it('should require preceeding /', function (done) {
           var cb = after(2, done)
           var router = new Router()
-          var route = router.route('/foo/*')
+          var route = router.route('/foo/(.*)')
           var server = createServer(router)
 
           route.all(sendParams)
@@ -770,23 +770,23 @@ describe('Router', function () {
         it('should work in a named parameter', function (done) {
           var cb = after(2, done)
           var router = new Router()
-          var route = router.route('/:foo(*)')
+          var route = router.route('/:foo(.*)')
           var server = createServer(router)
 
           route.all(sendParams)
 
           request(server)
           .get('/bar')
-          .expect(200, {'0': 'bar', 'foo': 'bar'}, cb)
+          .expect(200, {'foo': 'bar'}, cb)
 
           request(server)
           .get('/fizz/buzz')
-          .expect(200, {'0': 'fizz/buzz', 'foo': 'fizz/buzz'}, cb)
+          .expect(200, {'foo': 'fizz/buzz'}, cb)
         })
 
         it('should work before a named parameter', function (done) {
           var router = new Router()
-          var route = router.route('/*/user/:id')
+          var route = router.route('/(.*)/user/:id')
           var server = createServer(router)
 
           route.all(sendParams)
@@ -799,7 +799,7 @@ describe('Router', function () {
         it('should work within arrays', function (done) {
           var cb = after(3, done)
           var router = new Router()
-          var route = router.route(['/user/:id', '/foo/*', '/:action'])
+          var route = router.route(['/user/:id', '/foo/(.*)', '/:action'])
           var server = createServer(router)
 
           route.all(sendParams)
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -28,13 +28,27 @@ var hasOwnProperty = Object.prototype.hasOwnProperty
 
 module.exports = Layer
 
-function Layer(path, options, fn) {
+function Layer(p, options, fn) {
   if (!(this instanceof Layer)) {
-    return new Layer(path, options, fn)
+    return new Layer(p, options, fn)
   }
 
   debug('new %o', path)
   var opts = options || {}
+
+  // If not in strict allow both with or without trailing slash
+  var path = p
+  if (!opts.strict) {
+    if (!Array.isArray(path) && path !== '/' && path[path.length - 1] === '/') {
+      path = path.substr(0, path.length - 1)
+    } else {
+      for (var i = 0; i < path.length; i++) {
+        if (path[i] !== '/' && path[i][path[i].length - 1] === '/') {
+          path[i] = path[i].substr(0, path[i].length - 1)
+        }
+      }
+    }
+  }
 
   this.handle = fn
   this.name = fn.name || '<anonymous>'
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "debug": "3.1.0",
     "methods": "~1.1.2",
     "parseurl": "~1.3.2",
-    "path-to-regexp": "0.1.7",
+    "path-to-regexp": "^3.0.0",
     "setprototypeof": "1.1.0",
     "utils-merge": "1.0.1"
   },
--- a/test/req.params.js
+++ b/test/req.params.js
@@ -137,7 +137,7 @@ describe('req.params', function () {
           })
         })
 
-        router.get('/*', hitParams(1))
+        router.get('/(.*)', hitParams(1))
 
         request(server)
         .get('/buzz')
@@ -156,7 +156,7 @@ describe('req.params', function () {
           })
         })
 
-        router.get('/*', hitParams(1))
+        router.get('/(.*)', hitParams(1))
 
         request(server)
         .get('/bar')


dougwilson:
> Also. you will notice a change in the tests. I opened a separate issue to discuss, #41.

The test has been fixed for quite a while on master. I can see this PR cannot even be merged by GitHub, so I think you have it based off some stale code.

wesleytodd:
Yep, forgot to update my forked version.  Will do that now, sorry.

dougwilson:
And yes, bumping would be 2.0, which is fine :) Can we get a migration guide written up for users to understand how to translate their routes from the old style to the new style to add with this? I think that's really the only blocker.

wesleytodd:
Based on the tests there is one breaking use case because `*` is not a valid regexp.  This means `/:foo(*)` needs to become `/:foo(.*)`.

I think @blakeembrey could probably provide a more comprehensive list of changes, but then I would be happy to write them up into a guide.

blakeembrey:
I've written the differences a bundle of times, I think the most comprehensive was for Doug in the Express repo.

dougwilson:
The tests in this module do not actually exercise the different syntaxes currently, so knowing what broke is not possible from this test suite (for example, I don't see you adding any tests for all the new syntax in 1.x, continuing that trend ;) ).

wesleytodd:
I can add a bunch more tests for the supported syntax if you like.  It would seem weird to test features of `path-to-regexp` in this module though, so I would prefer to keep it the way it is.

For documentation, is what is [listed here](https://github.com/pillarjs/path-to-regexp#compatibility-with-express--4x) comprehensive?  Can we just do a link here to that section?

wesleytodd:
I cannot find any more docs that what is in that readme.  So I added them here.

As for tests, I think testing dependencies is not a great decision.  But I understand the reasoning here because `path-to-regexp` makes up a HUGE portion of the interface for the router.  If we could come up with a way to exercise the tests from the underlying module where when it updated, we automatically were testing the new features, then I think it would be fine.  Otherwise it is just a maintenance headache.

Either way, this update does not make us worse off.  So IMO it could be merged.

dougwilson:
I agree that extensively testing dependencies is not great. But sometimes I still think that at least basic tests of the main features of a dependency _as they are directly exposed through the interface of this module_ helps flush out accidental regressions upstream and can help us understand if we can bump a major in that dep without bumping our dep.

dougwilson:
Thanks for starting on the migration guide :) ! One thing I noticed right away is that the migration guides doesn't provide any guidance on whatever the reason the tests were changed in order to up this dep?

wesleytodd:
I added tests, I think they should be satisfactory, hits the major points.  

As for the guide thing, I am really unsure what the core changes were, so I don't actually know what to add.  I am happy to add what other people post here, or let them add to this PR as necessary.

gajus:
The latest version is 2.0.0.

dougwilson:
Yea, so we still need a migration guide from the respect of this module. Users need to understand what routes they need to change and how during the upgrade or they'll never be able to upgrade their apps successfully. It seems the 2.0 release breaks even more tha. The 1.0, which is fine, but I'm not even clear on how to upgrade the route paths for the transition.

blakeembrey:
Just a note, but compatibility is still listed in https://github.com/pillarjs/path-to-regexp#compatibility-with-express--4x. The primary difference between 1.x and 2.x was that I needed to drop support for `*` - it wasn't clear how it was meant to be encoded when converting a "path string" like `/*` into `/foo/bar`.

wesleytodd:
Ok, I just updated this PR.  But it appears that something also changed with the handling of trailing slashes.  @blakeembrey Know what I need to do to fix the two tests I added `skip` to?

test/route.js
```diff
@@ -633,23 +633,23 @@ describe('Router', function () {
...
-          var route = router.route('/:foo(*)')
+          var route = router.route('/:foo(.*)')
           var server = createServer(router)
 
           route.all(sendParams)
 
           request(server)
           .get('/bar')
-          .expect(200, {'0': 'bar', 'foo': 'bar'}, cb)
+          .expect(200, {'foo': 'bar'}, cb)
```
wesleytodd: Hmm, I did not do this change.  I wonder what I did.  I will look into why this is changed and fix it before we merge this.

wesleytodd:
I just updated the upgrade guide as well.  Any thoughts or changes necessary?

dougwilson:
Does this include the changes between 0.1.x and 2.x or just 0.2.x and 2.x? What were the differences between 0.1.x and 0.2.x, again?

wesleytodd:
AFAIK, this guide is ALL of the changes since `0.1.7`.

test/route.js
dougwilson: I'm not sure if we need to keep the old behavior if the new behavior is better, but we need to clearly explain how someone looking at their routers will know that the result of req.param is going to change

wesleytodd: Ok, I will add this to the guide and leave the change here.  I do think the named argument is what I would expect from the route defined in this test.

README.md
```diff
@@ -332,6 +332,22 @@ curl http://127.0.0.1:8080/such_path
 > such_path
 ```
 
+## Migrating to 2.x from 1.x
+
+The main change is the update to `path-to-regexp@2.0.0`, which has a few breaking changes:
+
+- No longer a direct conversion to a RegExp with sugar on top - it's a path matcher with named and unnamed matching groups
+  - It's unlikely you previously abused this feature, it's rare and you could always use a RegExp instead
```
dougwilson: We should explain how to know if you are using the feature rather than editorial that "you're probably not using it"

README.md
```diff
@@ -332,6 +332,22 @@ curl http://127.0.0.1:8080/such_path
...
+- All matching RegExp special characters can be used in a matching group. E.g. `:user(*)` needs to become `/:user(.*)`
+  - Other RegExp features are not support - no nested matching groups, non-capturing groups or look aheads
```
dougwilson: Typo in "support"

README.md
wesleytodd: @blakeembrey I copied this from your readme.  Can you give me a short blurb to help explain what this means?

README.md
wesleytodd: fixed

test/route.js
wesleytodd: done

README.md
blakeembrey: Sure. It's anywhere someone is abusing stuff like `/\\d+` outside of a matching group (e.g. instead of `/(\\d+)`). It's possible it's used in places today, but wouldn't be too useful since it wouldn't result in a match the user could use.

blakeembrey: If you end up coming up with a better or more succinct explanation, happy to replace the one in the README 😄

test/router.js
```diff
@@ -730,7 +730,7 @@ describe('Router', function () {
       .expect(200, 'saw POST /bar', cb)
     })
 
-    it('should match if path has trailing slash', function (done) {
+    it.skip('should match if path has trailing slash', function (done) {
```
blakeembrey: Ah, this one is an ongoing discussion but will likely be fixed in a 2.1 release, sorry - https://github.com/pillarjs/path-to-regexp/issues/123. The old one only passed because it always trimmed the trailing slash, the new one is leaving it there but I'm going to adjust the behaviour in non-ending mode to support this match behaviour.

README.md
dougwilson: You'd be surprised what people do with these routes. Express 4.16.1 is a release just because it seemed to be popular to do something thag shouldn't work, but did.

test/router.js
blakeembrey: @dougwilson I could use your input though, does this feel valid to you. Building up a path, from looking at it, should be `/foo//` and `/foo//foo`, but in this mode it acts differently to written. I'm interested in the viewpoint in support of this oddity.

dougwilson: I'll need to get a computer to understand the question, because mobile provides no way to view outside the diff to even see the path and url in the test.

Also thinking on this, i think we should just do a router 2 alpha now. Should we wait for 2.1.0 first or just put this pr back to 1.x and release a router 2 alpha 1 with it and follow with a alpha 2 after the 2.x regexp changes are settled?

README.md
wesleytodd: Question: What specifically does this mean?

> Parameters have suffixes that augment meaning - `*`, `+` and `?`. E.g. `/:user*`

Couldn't we always use those suffixes?  How are they different from `0.1.7` to `2.0.0`?

test/router.js
wesleytodd: I am available this weekend to finish up what we need to get this merged into a 2 alpha, so if we could do that it would be awesome.  IMO getting this sooner will help people adjust better and give them more time to work with the new ways.

dougwilson: Yea. That's why I'm wondering if we just start with the first alpha using path to regexp 1.x

wesleytodd: @blakeembrey How much is involved in putting together a PR for https://github.com/pillarjs/path-to-regexp/issues/123#issuecomment-335948313?  I haven't actually worked in `path-to-regxp` before, but if its simple enough I can take a stab at it before we decide what to do with this alpha.  If its too much, I can just revert this back to `1.7.0` simple enough.

README.md
blakeembrey: No, 0.x never had a concept of "suffixes". There's places where it happened to work because everything is treated as a regexp with based search and replace. E.g. `/*` became `/(.*)` in search and replace (that's the reason you can't use `*` in Express.js). It also means that, yes, stuff like `/:user?` might have worked but purely because of the search and replace implications and definitely not consistently (for instance, `/:test*` never would have done anything expected).

Note: I haven't tested the specific examples above, but it's possible a comparison (0.x -> 2.x) might be the easiest way to understand the differences.

test/router.js
blakeembrey: It should be very simple and I can implement it now, if that's what you need.

README.md
wesleytodd: Yeah that is probably why those worked for me :)

I was actually thinking it might be nice to update the demo app [1] to support multiple versions, which might be helpful for people to discover the differences for their routes.  Maybe I can work on that next.

[1]: http://forbeslindesay.github.io/express-route-tester/

test/router.js
blakeembrey: It's mostly been a matter of figuring out how it _should work_. Right now, based on feedback and these tests, I'm thinking (using `/foo/` input):

* `end: true, strict: false` - `/foo//?`
* `end: true, strict: true` - `/foo/`
* `end: false, strict: false` - `/foo(?=/)`
* `end: false, strict: true` - `/foo/`

README.md
wesleytodd: Ok, got a PR for that feature in the demo app: https://github.com/ForbesLindesay/express-route-tester/pull/11

test/router.js
blakeembrey: Actually, comparing this one to https://github.com/pillarjs/path-to-regexp/issues/123, it's slightly different. Supporting https://github.com/pillarjs/path-to-regexp/issues/123 makes sense and is pretty straightforward. The reasoning behind supporting this one is less so. I'll wait for @dougwilson for feedback on why this style is supported since the `.use` is identical as using `/foo` here (it's specifically testing an edge case for this test).

dougwilson: Ok, so I can see the entire test and it is testing that `app.use('/foo', fn)` runs `fn` for both `/foo` and `/foo/bar`. I think the reason this is tested and I continue to see many, many apps where they use the trailing slash in `app.use` is because it looks similar to `app.use('/', fn)` in that it ends with a slash. I'm not really sure which should be "right", but ideally with the amount of apps this would break, I wonder if it's worth all the turn over in everyone's apps.

Also: I'm describing the issue here correctly, right?

blakeembrey: Sounds correct, though I think you meant `app.use('/foo/', fn)`. With https://github.com/pillarjs/path-to-regexp/issues/123, it'll continue to work, but the child router won't pass in this example since it'll match as `/foo/` or `/` (whatever is input specifically) and the route trimmed will end up as `/foo/bar` -> `bar`. I definitely understand the potential churn here, but I've seen a few issues to the opposite effect which I'm trying to handle here - the unexpected behaviour of `'/'` becoming just `''` means you can't actually mount anything at specifically `/foo/` instead of `/foo`. Previous versions have gotten around the issue by saying "set `strict: true`", and if that's desired I can certainly do that.

README.md
wesleytodd: If you all need to see the differences in the versions you can view it here until the PR is merged: https://wesleytodd.github.io/express-route-tester/

test/router.js
wesleytodd: So are we just waiting to publish the `2.1.0` (or whatever you plan to version this as)?  Or is there anything else we need to get this pulled together?

dougwilson: Yea, it sounds like to bump this to at least 2.1.0 (or back to 1.x and other as 2.x). I can merge whichever into the 2.0 branch that is going on.

blakeembrey: @dougwilson How important is this given it's a backward incompatible release of `path-to-regexp` and, I believe, it's a clearer implementation today. Here's a couple of related discussions: https://github.com/pillarjs/path-to-regexp/issues/120 and https://github.com/pillarjs/path-to-regexp/issues/123. With this change, the features in these issues would not be possible (specifically the trailing slash enforcement in https://github.com/pillarjs/path-to-regexp/issues/123, which larger worked by accident in 1.x).

dougwilson: I mean, anything can change, I don't mind as long as we have migration instructions or at least a list of all the changes. Are those differences listed in the PR here already?

Of course Express continues to be num 1 framework for Node because the stability (in API, that is). The more differences the more it will invalidate all books and tutorials out there. It's not a blocker, only somethig to think about.

blakeembrey: Makes sense. I'd love to promote not using `app.use('/', fn)` as mentioned, since it's better overall just to skip the regexp (I believe this module handles it as a special case anyway). The only case it would break is the sub-router if it were trimming based on the matched `/`. However, even that I believe is handled with https://github.com/pillarjs/router/blob/ca2d51c145baa1bb4a3d4b55985b728560c328a1/index.js#L320. Let me do a 2.1 release and merge it into here, chances are it _just works_ because of these two handlers already built into the router that may have been handling 0.1.x edge-cases (for 1.x I had tried to bring that logic into `path-to-regexp`, for 2.x I realised pinning a ton of behaviour to Express.js edge cases might be a worse off idea).

blakeembrey: Done, `2.1.0` is released. When you're ready @wesleytodd, feel free to pull it in and see if it works as expected now. From looking at the module, I believe the edge-cases are all handled at the router level already.

wesleytodd:
Ok, updated!  Tests passed locally.

wesleytodd:
Ok, so I updated the tests to reflect the breaking change.  Before I write an addition to the change log in the readme, can you all take a look and ensure this is what we all agree we should do?  Specifically `use('/foo/' ...)` will no longer match `/foo` or `/foo/bar`.

You can check these routes with the new behavior of `2.1.0` at https://wesleytodd.github.io/express-route-tester/

test/router.js
```diff
@@ -742,12 +742,12 @@ describe('Router', function () {
...
 
       request(server)
-      .post('/foo')
+      .post('/foo/')
       .expect(200, 'saw POST /', cb)
 
       request(server)
       .post('/foo/bar')
-      .expect(200, 'saw POST /bar', cb)
+      .expect(404, cb)
```
dougwilson: I mean, there is no way this is actually the intended behavior, right? If `/foo/` is not a prefix of `/foo/bar` I don't know what `.use` is supposed to be doing.

wesleytodd: Well you you be required to do `.use('/foo')` If you wanted to match `/foo/bar`.

dougwilson: Yea, but then what will it match? If having the trailing / is not correct, then why not throw? I'm afraid this is going to be very confusing.

dougwilson: Using https://wesleytodd.github.io/express-route-tester/ I choose 2.1.0 and it says `/foo/` with no check boxes matches `/foo/bar`. It seems either that test is wrong or there is something else going on here that needs to be fixed.

wesleytodd: `use('/foo/'` will match `/foo/` and nothing else.  It is not that it is "not correct", it just has a very specific use case.  If you wanted the old behavior, all you have to do is change to `use('/foo'` I think.

wesleytodd: You need to have `end` checked to match what is un a use call I think:

```
var layer = new Layer(path, {
  sensitive: this.caseSensitive,
  strict: false,
  end: false
}, fn)
```

dougwilson: If end is checked, that means `end: true`, but like you posted, `end` is false, which is why  unchecked it. The `.use` is the in-ending paths.

wesleytodd: Yeah sorry, I should have stopped before posting that, deleted right after posting it lol.

dougwilson: haha. Cleaned up the convo :)

wesleytodd:
I was looking through the remaining list of things to do for express 5.0, and this is one of the slated changes.  After the last additions does anyone see any issues publishing this and moving forward on `router@2`?

test/router.js
dougwilson: That defeats the entire purpose of `.use`, does it not? How is `/foo/` not a prefix of `/foo/bar` ? Does `/foo/bar` not begin with `/foo/` ?

dougwilson:
I still think that `.use` is broken with this change, so I would not want to merge it yet.

test/router.js
wesleytodd: The point of `.use` is maintained without the trailing `/`, and I thought this was reasonable.  To get that behavior back we would have to do make an exception for the behavior of `path-to-regexp`.  AFAIK we only do that for `*` and `/`, adding this exception would be a bit more involved because it is any path with a trailing `/`, and possibly involve other edge cases.

Would it be alright if we directly documented this new behavior?  Or do you think working on this code to add this exception back at the router layer would be good?

dougwilson: I'm not sure. To me, `.use('/foo/', fn)` seems to cause completely unexpected behavior from how Express.js `.use` has always worked forever before. It's in many books and I think breaking this is going to be a large uphill battle from a user support perspective.

I think that either (a) we need to make it work as expected or (b) it shouldn't even be an accepted argument and throw / error out early.

I'm actually curious, perhaps before we proceed, that if you were to document this "new" behavior, what would it sound like / what would the use-case for this type of argument for `.use` be? It just seems really weird if `.use` works two completely different ways depending just on if you happen to add a trailing `/` on the string argument. Seems very unintuitive, but maybe I just don't understand the use-case this new behavior is enabling?

dougwilson: Also, perhaps the answer is that `path-to-regexp` no longer meets the needs of Express.js and we need to scrap it for something else. Just putting things on the table.

wesleytodd: > large uphill battle from a user support perspective

Yeah we should avoid that if reasonable.  FWIW, if I ever wanted to serve a path like this I never put a trailing `/`.  I am not sure how common either one is but every project I have worked on has not used the trailing `/`.



>  (b) it shouldn't even be an accepted argument and throw / error out early.

I think we need some version of the trailing slash if someone wants specifically to do something like serve a directory listing in this case.  So erroring is not really an option.

> I think that either (a) we need to make it work as expected

I am looking at what this would take.  I think it is more simple than I first thought.  See the new commit I added.  Is this satisfactory?

> Also, perhaps the answer is that path-to-regexp no longer meets the needs of Express.js

I would hope we can avoid this unless we actually wanted to change the routing/matching semantics significantly.

wesleytodd:
Ok, I added an exception so that when we are not in strict mode it will have the existing behavior.  Calls to `.use` are always non strict mode, so `use` works as before.  Does that work?

blakeembrey:
Would it help if I identified all the edge cases and the module issued deprecation warnings when encountered? I don’t think any of the cases, except the primary RegExp changes, should be relied upon anyway, but it should be simple to notify the user on how to handle it properly - would be a good opportunity to make sure all these edge cases are legitimate too or whether I’ve removed some useful functionality.

The deprecation warning could be skipped with an environment variable (or enabled with one as an “upgrade guide”).

blakeembrey:
Another idea is this module could accept functions as the path argument to easily allow people to switch in other path matching semantics. Upgrading to 2.x is probably in many people’s interests though, I know the latest versions are in use by a lot of front end frameworks that are likely trying to support isomorphic use cases and right now the paths don’t match up with Express.

wesleytodd:
> Would it help if I identified all the edge cases and the module issued deprecation warnings when encountered?

I don't think that would help all that much.  That could just as easily be done in express.  I think the issue here is how behind we are, and how difficult it has been to update because it was neglected for so long.  I think in the future this should be solved by updating more regularly.

> Another idea is this module could accept functions as the path argument to easily allow people to switch in other path matching semantics.

IMO the value is having a common set of understood semantics.  Having it be optionally different would mean we could not get the value out of the community agreement on these semantics.  Again I think pushing updates more regularly would be valuable :)

wesleytodd:
Just pushed an update to remove the other test changes.  It was failing for array of path conditions, but everything looks good now.  Are we good to merge this?

wesleytodd:
I rebased this and pointed it to the `2.0` branch, but noticed that 3.0 is out.  @blakeembrey should I update this to that?  What broke between 2.0 and 3?

blakeembrey:
@wesleytodd See https://github.com/pillarjs/path-to-regexp/blob/master/History.md#300--2019-01-13. Specifically it was the treatment around delimiter tokens (e.g. the character before `:token`), it was improved to handle a couple of use-cases more elegantly and removed "partial" token support which is actually better handled by escaping characters.

wesleytodd:
Awesome, so I updated to `3.0`, it caused one test failure.  I added another section to the readme to describe the change, and made the corresponding test change.  This should be good to merge now I think.

wesleytodd:
Too many versions too fast :P

@blakeembrey Should we update this to `5`? What are the breaking changes in 4 and 5?

blakeembrey:
@wesleytodd See https://github.com/pillarjs/path-to-regexp/releases. 4.0 was for ES2015, 5.0 was because I added a feature(s) I shouldn't have.

wesleytodd:
Ahh, I was wondering about that quick jump from 3 to 5.  There was too many notifications for me to read all of them and follow.  Could you summarize the changes so we can add it to the readme and docs?  I will update the PR from there.

blakeembrey:
@wesleytodd The only mentionable feature is support for nested non-capturing regex groups, e.g. `/(abc(?=d))`. No breaking changes in these releases for paths, only library usage (e.g. `module.exports` -> `module.exports.pathToRegexp`).

blakeembrey:
Final release I intend to make for this library is https://github.com/pillarjs/path-to-regexp/releases/tag/v6.0.0, which reverts some of the confusion I added in v3 and adds an alternative (`{}`) to support the reason I did the any character prefix in v3.

README.md
```diff
@@ -382,6 +382,64 @@ router.route('/pet/:id')
...
+router.get('/:foo(.*)')
+
+// old GET /bar
+console.log(req.params) // {0: 'bar', 'foo': 'bar'}
+
+// new GET /bar
+console.log(req.params) // {'foo': 'bar'}
+```
+
+#### Partial matching, prefer escaping delimiter
```
blakeembrey: This section shouldn't be relevant anymore. However, you can make the example `/user{s}?` if you wanted to.

test/route.js
```diff
@@ -645,7 +645,7 @@ describe('Router', function () {
         it('should work following a partial capture group', function (done) {
           var cb = after(2, done)
           var router = new Router()
-          var route = router.route('/user(s)?/:user/:op')
+          var route = router.route('\\/user(s)?/:user/:op')
```
blakeembrey: Should be good to revert this.

gireeshpunathil:
this needs a rebase. anything else?

dougwilson:
This needs nothing at this time as Wes rebased it on a separate branch instead of within the PR. Github will close it automatically after I move the changes back to the PR branch.

dougwilson:
So it looks like I don't have permission to push up to the remote branch. This is just a heads-up, so the changes will end up landing on the 2.0 branch, and will end up stating this PR is "closed" instead of "merged".

gireeshpunathil:
> the changes will end up landing on the 2.0 branch, and will end up stating this PR is "closed" instead of "merged".

@dougwilson - can you clarify what does this mean? the PR is originally destined for 2.0 branch itself, right? or is there a change? or is it that the landing will be same (to 2.0) but the github web interface will show it as closed, while the code would have landed?

dougwilson:
This PR has merge conflicts that must be resolved, and even has an incomplete section in the migration guide among other changes that are needed. I am working on the changes, but I get permission denied trying to push them up to the PR branch. Since I cannot push them into this PR first, landing this PR + changes onto 2.0 will cause GitHub to mark this PR as "closed" instead of "merged", is all.

dougwilson:
All the changes here + various fix ups has landed into the `2.0` branch 🎉

PR #47, opened 2016-08-14
Catches errors from Promises / Async functions

leebyron:
When returning Promises from a route handler either explicitly or implicitly by using Async Functions, default error handling is lost.

``` js
// Error caught and forwarded as expected
app.get('/', (req, res) => {
  throw new Error('Oh crap!')
})

// Error swallowed and lost
app.get('/', async (req, res) => {
  throw new Error('Oh crap!')
})
```

This PR began as https://github.com/expressjs/express/pull/3053, however it was a breaking change and Route/Layer have been moved to `router` for express@5.

### DIFF
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -66,7 +66,12 @@ Layer.prototype.handle_error = function handle_error(error, req, res, next) {
   }
 
   try {
-    fn(error, req, res, next)
+    var result = fn(error, req, res, next)
+    if (isPromise(result)) {
+      result.then(null, function (error) {
+        next(error || new Error('Promise rejected with: ' + error))
+      })
+    }
   } catch (err) {
     next(err)
   }
@@ -90,7 +95,12 @@ Layer.prototype.handle_request = function handle(req, res, next) {
   }
 
   try {
-    fn(req, res, next)
+    var result = fn(req, res, next)
+    if (isPromise(result)) {
+      result.then(null, function (error) {
+        next(error || new Error('Promise rejected with: ' + error))
+      })
+    }
   } catch (err) {
     next(err)
   }
@@ -172,4 +182,18 @@ function decode_param(val){
 
     throw err
   }
+}
+
+/**
+ * Returns true if the val is a Promise.
+ *
+ * @param {any} val
+ * @return {boolean}
+ * @private
+ */
+
+function isPromise(val) {
+  return val !== null &&
+    typeof val === 'object' &&
+    typeof val.then === 'function'
 }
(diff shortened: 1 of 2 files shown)


blakeembrey:
Duplicate of https://github.com/pillarjs/router/pull/32 and probably a couple of other PRs. If you're going to build this feature, remember that promises can be rejected with a falsy value and this doesn't handle it right now.

leebyron:
Updated to handle falsey rejection as well.

I really appreciate the work you put into #32, but my hope is that this is a minimal fix for this specific issue to incrementally introduce working with Promises into router/express while preserving as much existing behavior as possible to reduce migration pain.

As @dougwilson pointed out in expressjs/express#3053, with this PR the only major detail library authors will need to consider is that if they wrap middleware or route handlers that they return the result of calling the original middleware/handler.

leebyron:
> remember that promises can be rejected with a falsy value and this doesn't handle it right now.

This is a great point, and a similar issue currently exists for throwing falsey values which is something that's possible but pretty uncommon. Let me know if you would like me to open an issue for that or approach a fix.

leebyron:
Let me know if there are other concerns - I'm happy to add more tests or add other mitigations.

leebyron:
Just revisiting this since it's been a month or so - any evolution of thinking on this? Other concerns to address?

olalonde:
Bumppp...

gajus:
@dougwilson Is any help needed here?

jmar777:
I'm not sure if it's helpful or not, but I've been using an outside implementation of this in production for awhile now: https://github.com/jmar777/node-async-router. If nothing else, it might be useful to @olalonde, if you were looking for a similar solution.

It handles the error scenario mentioned by the OP ([relevant test case](https://github.com/jmar777/node-async-router/blob/master/test/router.verb.js#L66)).

wesleytodd:
LGTM :)

dougwilson:
This PR is going to land in the new 2.0 branch and release as an alpha version for preview this upcoming week.

dougwilson:
Sorry for the delay. I just rebased the `2.0` branch and now getting this rebased and up there. It looks like I don't have permission to push to the fork for this PR, so it won't show merged after I merge it, just "closed", but it'll be merged into `2.0`.

dougwilson:
Alright, I got this committed to the `2.0` branch now. Just checking a few more things, but otherwise will close this PR as merged at release the first 2.0 alpha 🎉

dougwilson:
Published to npm as version `2.0.0-alpha.1`

PR #49, opened 2016-10-24
Add Router events

hacksparrow:
The following events were added:
1. `handlestart` - emitted when the router starts to handle the router stack.
2. `layer` - emitted when a layer is matched by the router. It returns the matched layer.
3. `handleend` - emitted when the router has completed handling the router stack.

This feature should help developers handle requirements as described in https://github.com/expressjs/express/issues/2501, https://github.com/expressjs/express/pull/3100, and similar. 

Example of usage with Express:

```js
app.router.on('handlestart', function () {
  console.log('------- %d -------', Date.now())
})

app.router.on('layer', function (layer) {
  console.log(layer)
})

app.router.on('handleend', function () {
  console.log('------- %d -------', Date.now())
})
```

@dougwilson @danieljuhl @evanshortiss \o/

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,81 @@ curl http://127.0.0.1:8080/such_path
 > such_path
 ```
 
+## Events
+
+The router emits three events - `handlestart`, `layer`, and `handleend` - as it processes requests.
+
+### 1. `handlestart`
+
+This event is emitted when the router starts to process a request.
+
+Example:
+
+```js
+router.on('handlestart', function (req) {
+  req.id = Math.random().toString(36).slice(2)
+  console.log('HANDLESTART:', req.id)
+})
+```
+
+### 2. `layer`
+
+This event is emitted when a route layer is matched.
+
+Layers are matched till one of them sends a response back to the client.
+
+Example:
+
+```js
+router.on('layer', function (req, layer) {
+  console.log(layer)
+})
+```
+
+### 3. `handleend`
+
+This event is emitted when the router has finished processing a request and sent a response.
+
+Example:
+
+```js
+router.on('handleend', function (req) {
+  console.log('HANDLEEND:', req.id)
+})
+```
+
+Here is a complete example of using router events in an Express 5 app.
+
+```js
+var express = require('express')
+var app = express()
+
+app.use(function (req, res, next) {
+  next()
+})
+
+app.get('/', function (req, res) {
+  res.send('HELLO')
+})
+
+app.router.on('handlestart', function (req) {
+  req.id = Math.random().toString(36).slice(2)
+  console.log('HANDLESTART:', req.id)
+})
+
+app.router.on('layer', function (req, layer) {
+  if (!('layerCounter' in req)) req.layerCounter = 0
+  console.log('LAYER: %s -> %d', req.id, ++req.layerCounter)
+  // console.log(layer) // uncomment this to log the layer
+})
+
+app.router.on('handleend', function (req, router) {
+  console.log('HANDLEEND:', req.id)
+})
+
+app.listen(3000)
+```
+
 ## License
 
 [MIT](LICENSE)
--- a/index.js
+++ b/index.js
@@ -20,6 +20,7 @@ var mixin = require('utils-merge')
 var parseUrl = require('parseurl')
 var Route = require('./lib/route')
 var setPrototypeOf = require('setprototypeof')
+var EventEmitter = require('events').EventEmitter
 
 /**
  * Module variables.
@@ -63,6 +64,9 @@ function Router(options) {
   function router(req, res, next) {
     router.handle(req, res, next)
   }
+
+  // make Router an EventEmitter
+  mixin(this, EventEmitter.prototype, false)
 
   // inherit from the correct prototype
   setPrototypeOf(router, this)
@@ -164,6 +168,7 @@ Router.prototype.handle = function handle(req, res, callback) {
   var self = this
   var slashAdded = false
   var paramcalled = {}
+  var events = self._events
 
   // middleware and routes
   var stack = this.stack
@@ -185,6 +190,16 @@ Router.prototype.handle = function handle(req, res, callback) {
   // setup basic req values
   req.baseUrl = parentUrl
   req.originalUrl = req.originalUrl || req.url
+
+  // trigger the "beginning of route handling" event
+  if (events && 'handlestart' in events) self.emit('handlestart', req)
+
+  // trigger the "end of route handling" event
+  if (events && 'handleend' in events) {
+    res.once('finish', function () {
+      self.emit('handleend', req)
+    })
+  }
 
   next()
 
@@ -268,6 +283,9 @@ Router.prototype.handle = function handle(req, res, callback) {
     if (match !== true) {
       return done(layerError)
     }
+
+     // trigger the "layer found" event
+    if (events && 'layer' in events) self.emit('layer', req, layer)
 
     // store route for dispatch on change
     if (route) {
--- a/test/router.js
+++ b/test/router.js
@@ -3,6 +3,7 @@ var after = require('after')
 var methods = require('methods')
 var Router = require('..')
 var utils = require('./support/utils')
+var Layer = require('../lib/layer.js')
 
 var assert = utils.assert
 var createHitHandle = utils.createHitHandle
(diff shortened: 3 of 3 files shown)


evanshortiss:
@hacksparrow this is nice. I guess "layer" would be the event I'm looking for, but where can I associate that to the request itself here? For example, being able to do the below, would be nice, but needs _req_ to be injected:

```js
app.router.on('layer', function (layer, req) {
  req.matchedLayers = req.matchedLayers ? req.matchedLayers.concat([layer]) : [layer]
})
```

evanshortiss:
The end goal being that I can have a middleware like so:

```js
function routeLogger (req, res, next) {
  res.on('finish', function () {
    // write to analytics, or log route
    console.log(req.matchedLayers); // Maybe do some formatting here to get the actual paths joined
  });
}
```

dougwilson:
We will also need to discuss how the events are bubbled (or not). For example, would a user be required to attach an event listener on every single router (and of course, understand the list of all their routers) or should this use a form of event bubbling between routers, similar to the DOM?

hacksparrow:
@evanshortiss great suggestion. "handlestart" and "handleend" would need the request context too.

@dougwilson excellent feed back. I will implement a bubbling version as well for further thoughts on the feature.

hacksparrow:
@evanshortiss the event handlers have been updated to the following:

```
app.router.on('handlestart', function (req, router) {

})

app.router.on('layer', function (layer, req, router) {

})

app.router.on('handleend', function (req, router) {

})
```

evanshortiss:
That looks solid to me. Should provide plenty of context for introspection 👍

dougwilson:
@hacksparrow why add the `router` to the event? Isn't that available within the event handler as the context (`this`)?

hacksparrow:
@dougwilson err yes! I will remove it from the params. Thanks.

hacksparrow:
So, just for the record, the signature now is:

```
app.router.on('handlestart', function (req) {

})

app.router.on('layer', function (layer, req) {

})

app.router.on('handleend', function (req) {

})
```

index.js
```diff
@@ -186,6 +192,9 @@ Router.prototype.handle = function handle(req, res, callback) {
   req.baseUrl = parentUrl
   req.originalUrl = req.originalUrl || req.url
 
+  // trigger the "beginning of route handling" event
+  if (events && routerEvents[0] in events) self.emit('handlestart', req)
```
dougwilson: Is `routerEvents[0] in events` actually saving any time? I'm not sure if accessing `_events` and using `in` on it is really a good long term API to use with event emitters.

index.js
```diff
@@ -269,6 +278,9 @@ Router.prototype.handle = function handle(req, res, callback) {
       return done(layerError)
     }
 
+     // trigger the "layer found" event
+    if (events && routerEvents[1] in events) self.emit(routerEvents[1], layer, req)
```
dougwilson: If we swap `req` and `layer`, all three events will conveniently start with `req` as the first argument.

test/router.js
```diff
@@ -983,8 +985,117 @@ describe('Router', function () {
...
+
+  describe('events', function () {
+
+    describe('"handlestart"', function () {
+      it('should pass the request object', function (done) {
+        var router = new Router()
+        var server = createServer(router)
+
+        router.on('handlestart', function (req) {
+          assert(IncomingMessage.prototype.isPrototypeOf(req))
```
dougwilson: I would just check for certain properties, perhaps add a request header and try to read it here, rather than checking the prototype specifically. I know there are a bunch of people interested in better browser support, so if there is no reason to hurt the support, I would err on the side of that.

test/router.js
```diff
@@ -983,8 +985,117 @@ describe('Router', function () {
...
+        var router = new Router()
+        var server = createServer(router)
+
+        var handlers = [passThrough, passThrough, passThrough]
+        router.use(handlers)
+
+        var counter = 0
+        router.on('layer', function (layer, req) {
+          counter++
+          if (handlers.length === counter) done()
```
dougwilson: I'd recommend using the `after` function for this type of tracking, which will throw when invoked too much.

dougwilson:
Does anyone have any thoughts on the event names? I'm neutral on them :)

crandmck:
Please add some docs to the README.

index.js
hacksparrow: Agreed. I decided on this to keep the naming convention "flexible". Will come up with an better alternative.

index.js
hacksparrow: 👍

test/router.js
hacksparrow: 👍

test/router.js
hacksparrow: 👍

index.js
evanshortiss: This is a good suggestion 👍

hacksparrow:
@dougwilson @crandmck I have updated the PR according to your suggestions.

For the record, the signature of the `layer` event listener now is:

```
app.router.on('layer', function (req, layer) {

})
```

index.js
```diff
@@ -186,6 +191,16 @@ Router.prototype.handle = function handle(req, res, callback) {
   req.baseUrl = parentUrl
   req.originalUrl = req.originalUrl || req.url
 
+  // trigger the "beginning of route handling" event
+  if (events && 'handlestart' in events) self.emit('handlestart', req)
+
+  // trigger the "end of route handling" event
+  if (events && 'handleend' in events) {
```
hacksparrow: @dougwilson what's your opinion of this implementation?

crandmck:
Thanks, LGTM.

wesleytodd:
Anything blocking this?  I would love to replace what I have been doing with middleware these new events.

dougwilson:
There are still issues I wrote above unaddressed, with the biggest that it is reaching into the internals of Node.js events which is unlikely to keep working in the future.

index.js
```diff
@@ -186,6 +191,16 @@ Router.prototype.handle = function handle(req, res, callback) {
...
+    res.once('finish', function () {
```
dougwilson: What is the purpose of adding the finish event listener? Especially because there are no tests around all the edge cases this will cause (for example, are we saying that if the response is already finished before the handleend, our events is not supposed to fire?)

index.js
```diff
@@ -164,6 +168,7 @@ Router.prototype.handle = function handle(req, res, callback) {
   var self = this
   var slashAdded = false
   var paramcalled = {}
+  var events = self._events
```
dougwilson: Let's not peak into the internals of Node.js if it's not required, and I see no reason this is required. If there is, please enlighten me :)

index.js
```diff
@@ -20,6 +20,7 @@ var mixin = require('utils-merge')
 var parseUrl = require('parseurl')
 var Route = require('./lib/route')
 var setPrototypeOf = require('setprototypeof')
+var EventEmitter = require('events').EventEmitter
```
dougwilson: This should probably be alphabetical like the rest.

wesleytodd:
Ahh, sorry @dougwilson I didn't open the "outdated diffs".  I completely agree with the comments, and am anxiously awaiting this merge :)  Thanks for the good work @hacksparrow!!

README.md
```diff
@@ -295,6 +295,81 @@ curl http://127.0.0.1:8080/such_path
...
+```js
+router.on('handlestart', function (req) {
+  req.id = Math.random().toString(36).slice(2)
+  console.log('HANDLESTART:', req.id)
+})
+```
+
+### 2. `layer`
+
+This event is emitted when a route layer is matched.
```
dougwilson: Can you define "route layer" in the docs?

README.md
```diff
@@ -295,6 +295,81 @@ curl http://127.0.0.1:8080/such_path
...
+
+Example:
+
+```js
+router.on('handleend', function (req) {
+  console.log('HANDLEEND:', req.id)
+})
+```
+
+Here is a complete example of using router events in an Express 5 app.
```
dougwilson: This should not be part of the handleend section.

README.md
```diff
@@ -295,6 +295,81 @@ curl http://127.0.0.1:8080/such_path
...
+
+```js
+var express = require('express')
```
dougwilson: Can the example be a real example instead of just a bunch of console log lines? I.e. demonstrate the use case people will use this for. People love to copy paste, and this is not something they would do that on.

test/router.js
```diff
@@ -983,8 +984,153 @@ describe('Router', function () {
       .expect(200, 'saw GET /bar', done)
     })
   })
+
+  describe('events', function () {
+
```
dougwilson: These tests have a very inconsistent style from all other tests if you can clean them up at some point :)

dougwilson:
Also, what ever happened with the whole event bubbling discussion? I feel like I've mentioned it before, and @hacksparrow said he would look into it, but never really anything seemed to happen there. So what is the thoughts on that? Does that make it easier to accomplish whatever the goals of this PR are or not?

hacksparrow:
@dougwilson I haven't made significant progress on event bubbling yet. Will update once done. In the mean time, it'd be great to hear the community's opinion about a bubbling and non-bubbling version.

evanshortiss:
@dougwilson @hacksparrow just to be clear, by bubbling you mean bubble to an upper "app" (express) instance? So for submounted apps, the event will bubble to the parent app and so on?

hacksparrow:
@evanshortiss bubbling within the routers (independent or the one that comes with Express) hierarchy. We were thinking along lines of the browser DOM events.

wesleytodd:
TBH I think you have to do bubbling for `layer` events, if you don't I think the behavior will be  unpredictable.  That being said, multiple start and end events seem problematic, or atleast overly complex.  I see two options on that:

1. "Bubbled" events get prefixed, or a different name altogether
2.  `handlestart`/`handleend` events don't bubble, but layer events still do

wesleytodd:
To address the cases that I would use this for right away, here are two I can think of:

1. Unbind events on the front-end when the NEXT `handlestart` event happens
2. Monitoring performance metrics for each layer

So for number one, having multiple start events for one "run" of the router stack, would mean I would have to track it myself.  Possible, obviously, but it would be nicer if the design compensated for this and gave me a simple way.

For number two, each "layer", IMO should include all sub routers.  Thus requiring bubbling.

wesleytodd:
So I am circling back on this, and after re-thinking it I do not think anything other than the `layer` event should be included.  Here are my thoughts:

1. There are complexities this introduced, as discussed above in detail.  None of which have simple resolutions.
  - Bubbling: the semantics of this are hard to get right for start and end, but pretty clear for layer
  - Handle End: Currently it is the response end, but this can already be handled with the `on-finished` package, and the router end has little to no value.
  - 
2. The feature's we have all mentioned can be solved in other ways:
  - Original Route/Matched Route: #34 solves this in a much better way
  - Metrics: If you actually care to get accurate metrics you should be capturing the start time sooner than this, and as mentioned above, `handleend` is an alias for `onFiished(res, cb)`.
  - Logging: Because #34 gets you all the info you need, you can then just log in a middleware at the end of your stack

That all being said, the one thing we cannot do with ease is track stats and metrics on individual layers.  Which is why I propose that this change just to:

- `layerstart`: Same as the current `layer` event is now
- `layerend`: Added once next is called after matching a layer

So I created #56 to illustrate and keep the discussion going.

fcolella:
Hi guys! so is this still happening? :)

wesleytodd:
> Hi guys! so is this still happening? :)

If it is, as I said in my last comment, I would hope we just go with the layer events.  That being said, I am not sure even this is the *best* way forward, and I have an idea which has been bubbling about adding a decorator mechanism which would allow for much more flexibility and also superseded the need to add these events at all.

I have left open my other PR since it might be good, but I think this PR should be closed.  If no one disagrees (@dougwilson or @hacksparrow) then I will close this in a week (reminder already set to revisit).

wesleytodd:
Ok, no comments.  But @dougwilson it looks like I am not an owner on this org so cannot close this. 🤷‍♂️

wesleytodd:
@dougwilson It looks like just adding me to the tc group is not enough to be able to close issues.  Seems like there must be some settings to make that work, but I am not sure.

dougwilson:
Sorry @wesleytodd about that. I will check it out and make sure your permissions are in order. You should have the ability do to this, so I will check why here shortly 👍

wesleytodd:
Finally coming back around here. This is superseded by https://github.com/pillarjs/router/pull/56 IIRC. Closing this, and assessing if that could possibly land for `router@2` and `express@5`.

Issue #62, opened 2017-11-21
Allow for identifying 404 vs. 405


whitlockjc:
Whenever `router` fails to match a request, it can be one of two reasons:

1. Method mismatch _(`405`)_
2. Path mismatch _(`404`)_

I cannot seem to figure out a way to differentiate between the two in my `finalhandler` and I was hoping that either someone can point me in the right direction or maybe `router` can be updated to allow for such things.

wesleytodd:
AFAIK the router package does not support this.  Looking through the code I think it would drastically change the way things are done to be able to support it.  Right now, if the method does not match, then it actually skips the whole path matching.  This is a perf improvement because the path matching is the least efficient part.

I *think* (although have not tried) that you could rig up your routes in such a way as you can do this manually, like:

```javascript
router.route('/foo')
  .get(getHandler)
  .post(postHandler)
  .all(function (req, res) {
    res.status(405);
  })
;
```

But you would have to be careful that nothing else after was supposed to handle the route.

whitlockjc:
Why does it have to be a drastic change?  Looking [at the source](https://github.com/pillarjs/router/blob/master/lib/route.js#L138), it seems very simple.  You could add a property to `err` based on whether the miss was due to method and/or path and let the `finalhandler` use this property to return a `404` or a `405`.  In fact, I'll submit a PR for this.

whitlockjc:
I am aware now that this would be a breaking change and that isn't ideal.  And since the test suite relies heavily on the current support...this could be bigger than expected.  But the logic for this seems pretty simple.

wesleytodd:
I didn't mean the code change was drastic, rather that the effect of the change would be.  Since it is quick I don't see a reason not to submit a PR, but I am not sure I am in support of testing route matching when the method does not match.

whitlockjc:
I thought about wrapping this up in a configuration option but after implementing this, there is no way to make it work.  For a method/path miss, there is no `err` sent to the `done` callback right now.  I _could_ add one for this but that would change the current handling so using a configuration option alone isn't enough.  I am not sure sending an `err` downstream is such a bad idea but it is a deviation from the current functionality and my concern would be breaking backward compatibility.

I'll leave this open for now but if you think it's not worth considering further, I'm cool if you close it.  I do think it's a good idea to be able to differentiate between `404` and `405`.

wesleytodd:
> I do think it's a good idea to be able to differentiate between 404 and 405

On this I agree completely.

[...]

PR #63, opened 2017-11-22
Added example of 404/405 handling as discussed in #62

wesleytodd:
If this in fact works then it is a nice addition since it is a bit more complex than the existing examples in the readme.  @whitlockjc Can you update if the solution actually solves your problem?

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,47 @@ curl http://127.0.0.1:8080/such_path
 > such_path
 ```
 
+### Example of advanced `.route()` usage
+
+This example shows how to implement routes where there is a custom
+handler to execute when the path matched, but no methods matched.
+Without any special handling, this would be treated as just a
+generic non-match by `router` (which typically results in a 404),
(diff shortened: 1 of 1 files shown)


whitlockjc:
If I understand this example, I would need to use `.all` for each path.  Is that correct?

wesleytodd:
Yes, that is correct.  I would say you might want to write a little helper to wrap it up:

var route = router.route('/foo')
// add methods...
handle405(route);

whitlockjc:
Yep, that's what I'd do.  Let me give it a shot.

wesleytodd:
Anyone have input on adding examples like this?  @whitlockjc said that the approach worked for his use case.  If no one opposed it would be nice to get this merged?

whitlockjc:
I wish there was a better way to do this, like a single `.all` for all routes.  But this does work as-is, verified.

wesleytodd:
>  not use Express-only syntax

Good call, I will remember this for the future.

PR #65, opened 2018-03-30
Handling promises at router layer

davidbayo10:
This feature will avoid try/catch block in all express controllers with async functions (async/await) or functions that returns a promise

### DIFF
--- a/test/router.js
+++ b/test/router.js
@@ -124,6 +124,92 @@ describe('Router', function () {
       .expect(404, cb)
     })
 
+    it('should support promise as route handler', function (done) {
+      if (!global.Promise) {
+        return done();
+      }
+
+      var router = new Router()
+      var server = createServer(router)
+
+      router.all('/:thing', function (req, res) {
+        var promise = new Promise(function (resolve, reject) {
+          setTimeout(function () {
+            resolve()
+          }, 100);
+        });
+
+        promise
+          .then(function () {
+            saw(req, res)
+          })
+
+        return promise
+      })
+
+      router.use(function (err, req, res, next) {
+        res.sendStatus(500)
+      })
+
+      request(server)
+        .get('/foo')
+        .expect(200, 'saw GET /foo', done)
+    })
+
+    it('should support catch rejected promise', function (done) {
+      if (!global.Promise) {
+        return done();
+      }
+
+      var router = new Router()
+      var server = createServer(router)
+
+      router.all('/:thing', function (req, res) {
+        var promise = new Promise(function (resolve, reject) {
+          setTimeout(function () {
+            reject();
+          }, 100);
+        });
+
+        return promise
+      })
+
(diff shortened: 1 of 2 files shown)


dougwilson:
Is this a duplicate of #47 ?

davidbayo10:
Yes, I did not see it. Close if you want or tell me what PR is better in terms of code.

dougwilson:
Ok. I'll review through your code to understand what it does. If you can provide highlights of the differences between the two PRs, that would be super helpful as well 👍 On the surface it seems the main difference is that #47 passes CI and this PR doesn't pass the CI.

davidbayo10:
I think my code is simpliest and more readable. That's true, I have not passed the CI. But I don't know why, because I passed tests (`npm test`) before push. 😔

dougwilson:
What is the behavior of this code for a promise that is rejected without a value (a valid rejection)? i.e. calling `reject()`

davidbayo10:
Actually, an incorrect behavior. Fixing...

davidbayo10:
image

This error comes from 2.0 branch. That's why builds are failing.

test/router.js
```diff
@@ -124,6 +124,92 @@ describe('Router', function () {
...
+        promise
+          .then(function () {
+            saw(req, res)
+          })
+
+        return promise
+      })
+
+      router.use(function (err, req, res, next) {
+        res.sendStatus(500)
```
dougwilson: There is no such thing as res.sendStatus

test/router.js
```diff
@@ -124,6 +124,92 @@ describe('Router', function () {
...
+          setTimeout(function () {
+            reject();
+          }, 100);
+        });
+
+        return promise
+      })
+
+      router.use(function (err, req, res, next) {
+        res.sendStatus(500)
```
dougwilson: There is no such thing as res.sendStatus

test/router.js
```diff
@@ -124,6 +124,92 @@ describe('Router', function () {
       .expect(404, cb)
     })
 
+    it('should support promise as route handler', function (done) {
```
dougwilson: What does this test? I ran this test without any of your changes and it passed. This test doesn't seem to actually depend on any change, or maybe I'm just missing something.

test/router.js
```diff
@@ -124,6 +124,92 @@ describe('Router', function () {
...
+          setTimeout(function () {
+            reject(new Error('async request failed'));
+          }, 100);
+        });
+
+        return promise
+      })
+
+      router.use(function (err, req, res, next) {
+        res.sendStatus(500)
```
dougwilson: There is no such thing as res.sendStatus

davidbayo10:
Ok, I think both PR have same functionality then Imy PR doesn't add value. As you said, the other author have spent his time in development. I expect that you release 2.0 version soon. Thank you for all. :)

PR #67, opened 2018-05-26
Mark 'options' as optional in JSDoc for Router

ankon:
The code handles them as optional with a default of '{}', and the README similarly shows the examples without any options argument.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+==========
+
+  * Fix JSDoc for `Router` constructor
+
 1.3.2 / 2017-09-24
 ==================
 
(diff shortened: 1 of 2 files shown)


ankon:
I've added some 'options' as a description of the option. Reads a bit funny, but avoids more warnings with linters that want no empty doc comments :)

dougwilson:
Would that be linters that run on user's of `router` projects, or run against the `router` source code itself?

ankon:
Typically you'd be running linters on your own code only, not the dependencies. ... Except for the linter in my head, which got quite conditioned already to fix empty JSDocs parts. I'm not sure about this project our your environment, but I can see defaults of users that open this one in their IDE make this appear as a warning.

In any case, I just wanted to explain why I added the word 'options' :)

dougwilson:
Gotcha. Yea, I was just checking if it's something external users would run into and so something we'd want to add a check to prevent in the future. It doesn't sound like this is the case here, though.

PR #68, opened 2018-07-01
add support for async/await

howtoclient:
added test if passed function is `async` function.
if it is then capture its error through `catch()`

### DIFF
--- a/test/route.js
+++ b/test/route.js
@@ -302,6 +302,40 @@ describe('Router', function () {
           .expect(500, 'caught: Bad promise rejection : null', done)
       })
 
+      it('should handle Promise null rejection in errors', function (done) {
+        var router = new Router()
+        var route = router.route('/foo')
+        var server = createServer(router)
+
+        route.all(function createError(req, res, next) {
+          next(new Error('boom!'))
+        })
+
+        route.all(helloWorld)
+
+        route.all(function handleError(err, req, res, next) {
+          var handler = function () {};
+          var catchFunction = function (errorHandler) {
+            handler = errorHandler;
+          };
+          setTimeout(function () {
+            handler(null)
+          }, 1)
+          return {
+            catch: catchFunction
+          }
+        })
+
+        route.all(function handleError(err, req, res, next) {
+          res.statusCode = 500
+          res.end('caught: ' + err.message)
+        })
+
+        request(server)
+          .get('/foo')
+          .expect(500, 'caught: Bad promise rejection : null', done)
+      })
+
       it('should handle errors from next(err)', function (done) {
         var router = new Router()
         var route = router.route('/foo')


howtoclient:
the tests are a problem. to get 100% coverage i need this to run on NodeJs 7.6 or higher. Witch means that eather it fails on lower versions but gets better coverage on the newer version or... i can mock the constructor... hmm

dougwilson:
So on the 2.0 branch we started dropping support for old Node.js versions (and can continue if necessary). We also have promise support PRs that will be merged into the 2.0 branch (see the 2.0 PR for the list). Wouldn't that PR for promise support automatically support async functions? Or I could be missing something on the difference since I'm not 100% familiar.

howtoclient:
it will, yes. but Promise relies on return.
i can use the 2nd version from before
`var promise=fn(req,res,next);`
`promise && promise.catch && promise.catch(next)`

howtoclient:
i am copying tests that use helloWorld, everything it does, async should do as well.

howtoclient:
There is an issue but i see its from 2016... 

https://github.com/pillarjs/router/pull/47

dougwilson:
Yea, #47 is what we are going to merge.

howtoclient:
lol... well, at-least i will finish what i started :)

howtoclient:
you have to admin, this one is a lot more elegant :)

howtoclient:
also you have to remove `README.md` from the test files...

dougwilson:
> you have to admit, this one is a lot more elegant :)

It may be, but it is missing the edge cases that #47 solved. For example a promise can reject the value null. Your PR here will incorrectly continue on to the next route instead of returning a failure 500 due to a promise rejection. Can you add all the edge cases to your PR we discussed and resolved in #47 ?


> also you have to remove README.md from the test files...

I'm not sure what you mean. We do validate that the code blocks in the readme have correct syntax in the CI if that is what you're referring to. We don't want to remove that check because the code blocks in the README need to be valid javascript.

howtoclient:
for example
`/home/travis/build/pillarjs/router/README.md
  31:7  error  Parsing error: Unexpected token function`
this is referring to: 
`async function getHelloWorld()=>{` 

the large nodejs support means i cant give usage example on `async/await` and default to something like `new Promise`

howtoclient:
can you explain what do you mean by
`For example a promise can reject the value null` ?

blakeembrey:
@howtoclient It's valid JavaScript to write `Promise.reject(null)` and have your catch method proceed to call `next(null)` which looks like no error occurred.

dougwilson:
> the large nodejs support means i cant give usage example on async/await and default to something like new Promise

You could always open the PR against the 2.0 branch in which case we have the ability to drop versions since it is a major release. At the very least I suspect you just need to change the syntax checker to accept async however you do that

> can you explain what do you mean by
For example a promise can reject the value null ?

The following doesn't work as expected in this PR, but works as expected in #47 :

```
app.use((req, res) => new Promise(resolve, reject) => {
  reject(null)
})
app.use((req, res) => res.end('this should not execute since promise above was rejected'))
```

howtoclient:
what should i do about the `README.md` situation?

dougwilson:
> what should i do about the README.md situation?

There is likely a setting in the syntax checker to specify what level of javascript syntax to validate against. I've never tried to put async in the readme before so don't know how. That would be an awesome contribution if you can figure out how to do that since we'll want to put async functions in the readme.

howtoclient:
i'll do you one better and allow ES 2017 :)

dougwilson:
Awesome. Make sure that is confined only to the README syntax check and not against the .js files, since they need to work in Node.js 0.8 still since this PR is against master :+1:

howtoclient:
there are 2 config files:
one for the js: `.eslintrc`
and one for .md files: `.eslint2017`

so you can edit them as you like.
https://eslint.org/docs/user-guide/configuring#example-configuration

howtoclient:
ok i think i'm done. :)

wesleytodd:
Hey @howtoclient, thanks for the contribution, but I am with the others that #47 is the way we should go.  In the future it will save you work if you start by searching or posting an issue asking about the feature.  In this case, the decision had already been made that #47 was moving forward, so it would have prevented you from wasting time.  We are super happy to welcome new contributors, and don't want to discourage you from making more contributions, so keep it up!

howtoclient:
Absolutely don't mind. Also if possible don't let PRs sit for 2 years...
Don't forget to copy the eslint config...

dougwilson:
Support now available in version `2.0.0-alpha.1`

DJviolin:
@dougwilson Can I use this router version `2.0.0-alpha.1` with express `4.17.1` or it will break?

Or I should use this package instead for async/await support:
https://github.com/express-promise-router/express-promise-router

dougwilson:
Hi @DJviolin yes, you can use any router with Express, even the 2.0.0-alpha.1 version of this module. Just `app.use(router)` where `router` is an instance from this module with your routes on it.

DJviolin:
I following the project structure created by `express-generator` and I placed this code in my `./app.mjs`:

```js
import Router from 'router';
// ...
const router = Router();
app.use(router);
```

So if I creating new routes, for example in `./routes/index.mjs`:

```js
import express from 'express';
const router = express.Router();
// ...
```

All of my routes will use router version `2.0.0-alpha.1` instead of the built-in one, correct?

dougwilson:
@DJviolin , no as `express.Router()` is the old router. You need to `const router = Router()` from this module in your `./routes/index.mj` file if you want to use this router.

Rather than continuing to comment in a years-old pull request (especially since no one else cal really see this and help you out), please jump into the Gitter channel for Express (or something else from the list):  http://expressjs.com/en/resources/community.html

Issue #48, opened 2016-10-14
Typescript Typings


origin1tech:
Any chance of getting some @types/router typings published? Or are there any floating around that we can make official and publish.

blakeembrey:
`@types` has no relation to this module, if you want someone in the community to type it you can ask at https://github.com/DefinitelyTyped/DefinitelyTyped which is where `@types` is published from. You can also contribute them yourself. I'll leave this issue open to add a TypeScript definition to the router itself, but it sounds like you're after something else that doesn't exist (yet).

origin1tech:
actually no I'm talking about the type definitions. However it would be appropriate to publish once done to @types for ease of install with Typescript 2.x. Sorry if I wasn't clear, but yes we'd need the defs first obviously :)

dougwilson:
@origin1tech I'm not sure if third parties even have the rights to publish under other user's namespaces, like the `@types` namespace. I think @blakeembrey has the right path forward, unless you can share some documentation on how we would be able to publish under that npm namespace?

I personally have never written typings before, so probably someone who knew how to write them would have to (I think @blakeembrey does).

blakeembrey:
@origin1tech I do know what you're talking about. That is the location you should ask if you want something published to `@types`. The `@types` on NPM has **zero** relation to the module itself. I would hopefully know, I am a member of @DefinitelyTyped, wrote @types and @typings and helped TypeScript formalize the `@type` implementation 😄

blakeembrey:
@dougwilson We can't, and we wouldn't want to. Once someone has written a definition, we can add it to the module ourselves and publish to NPM directly (using `typings` in `package.json` like https://github.com/pillarjs/path-to-regexp/blob/master/package.json#L6). This makes it easier for TypeScript users as they can forgo the installation from `@types` altogether and it _just works_.

origin1tech:
@blakeembrey I'm fully aware of that....my point is simply that it would be nice to get defs for router and then publish an @types project for ease of install. I get that they are unrelated, however if the author creates the definitions it's helpful as there's less chance of things getting missed.

blakeembrey:
@origin1tech You're missing the point. Please re-read my comments. If the goal is "ease of install" we would add them to **this module**, and not @types. That way, as a TypeScript user, you don't have to do **anything** to have it work - which is ease of install to me vs having to install an additional module.

origin1tech:
[...]

PR #76, opened 2019-01-08
Added typescript definition files

ilikejames:
Hand rolled types, though many are copied from the base type definition within [`express-serve-static-core`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/express-serve-static-core).

A few things to point out:
1. using Router without new will cause a type error. 
2. `Request` interface used in express is specific to express. I could have used node's `http.IncomingMessage` type, but I felt having this be isomorphic would be better as this is how I'd be using this. I think the code could be changed (non-breaking) to support different run-time environments better using generics but thats something i'll have to think about. In the meantime, this parameter should be cast. 
3. `Response` as above. 

I've added a test for the typechecking. Hopefully this will catch any deviations going forward. 

Ref #48


ilikejames:
Looks like typescript (for the test) requires `"node": ">=4.2.0"`, though it seems to be building for versions >=0.12. Happy to remove?

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-  - "0.8"
-  - "0.10"
```
dougwilson: Please add these back; this version of the module supports these Node.js versions. Removing support is a major version bump. You can retarget your PR to the 2.0 branch if you want to make a breaking change, though.

index.d.ts
```diff
@@ -0,0 +1,138 @@
...
+     */
+    all: IRouterMatcher<this>;
+
+    get: IRouterMatcher<this>;
+    post: IRouterMatcher<this>;
+    put: IRouterMatcher<this>;
+    delete: IRouterMatcher<this>;
+    patch: IRouterMatcher<this>;
+    options: IRouterMatcher<this>;
+    head: IRouterMatcher<this>;
```
dougwilson: There seems to be a lot more methods missing here. Can you add the rest? Examples are `.bind`, `.propfind`, etc.

package.json
```diff
@@ -18,28 +18,34 @@
     "utils-merge": "1.0.1"
   },
   "devDependencies": {
+    "@types/node": "^10.12.18",
```
dougwilson: So this is just a general question: will this cause issues for folks using this module with other Node.js versions like the 8.x series? If the types are for the 10.x series?

package.json
```diff
@@ -18,28 +18,34 @@
...
   "files": [
     "lib/",
     "LICENSE",
     "HISTORY.md",
     "README.md",
-    "index.js"
+    "index.js",
+    "index.d.ts"
   ],
+  "main": "index.js",
```
dougwilson: Please remove the `main` prop; it's not needed.

package.json
ilikejames: Its just for the typings on the test. It won't explode downstream.

.travis.yml
ilikejames: I'll retarget as typescript won't work with such old versions of node. 

I'll also thinking that this maybe better pushed to @DefinitelyTyped. This way there would be locked @typings/router@1.3.3 version aligned with a publish router@1.3.3 and leave it to the community to update. I often use older package versions simply due to the @type support. 

Looking down the list of PRs for 2.0 I not sure what methods we could do to keep these aligned, except by being rigorous for any forthcoming changes. 

Up to you guys really. Let me know.

index.d.ts
ilikejames: `.bind` and all the default object level props you get for free. I can't find `.propfind` in the codebase. Not on `2.0` either?

.travis.yml
ilikejames: Looks like I can roll typescript back to 1.8.0 and it should work with node >= 0.8.0

ilikejames:
Ah, seems I accidentally closed this! And the reopen button is greyed out. 
I'll open a new PR. This will be targeted at master.

package.json
ilikejames: done

index.d.ts
dougwilson: All these methods are not in the code base -- they are dynamic depending on the Node.js version -- even the ones like `.options` is not in the code base at all...

We iterate over https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_http_methods and add the lower case as a method to the router prototype. Here is propfind existing:

```
$ node -pe 'require("router")().propfind'
[Function]
```

PR #81, opened 2019-07-18
deps: setprototypeof@1.2.0

wesleytodd:
Doesn't have impact here, but this release includes mitigation for a possible prototype pollution.  I will be opening up a series of these on the Express repos.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,7 +3,7 @@ unreleased
 
   * deps: array-flatten@2.1.2
   * deps: parseurl@~1.3.3
-  * deps: setprototypeof@1.1.1
+  * deps: setprototypeof@1.2.0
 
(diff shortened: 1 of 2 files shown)


wesleytodd:
Also of note, this can be ported directly onto the `2.0` branch.  If you give me a LGTM I will go ahead and do the merge and push, if you want to get me publishing things I can go ahead with a publish as well.  Maybe including #42 as that is ready.

wesleytodd:
Sorry!  I had left it because I just amended the commit in the `2.0.0-alpha.2` branch, but it makes sense like you said to merge here and rebase that.

dougwilson:
yea, otherwise we'll just have to end up cherry-picking it backwards instead of merge forwards is the only diff. it's no prob

PR #60, opened 2017-10-14
Release 2.0

dougwilson:
This is a tracking issue for release 2.0.

**_Try the latest beta with `npm install router@2.0.0-beta.1`_**

**Please keep feature requests in their own issues**

If you want to make a comment on a particular change, please make the comment in the "Files changed" tab so comments are not lost during a rebase.

List of changes for release:
- [x] Add basic support for returned, rejected Promises #47 
- [x] Add promise support to `router.param` #92
- [x] Drop support for Node.js below 0.10
- [x] Remove `debug` dependency
- [x] Upgrade `path-to-regexp` to 3.x #42 
- [x] Upgrade `path-to-regexp` to 8.x (#117) 

List of things to refactor:
- [x] Use standard style

**Testing this release**

If you want to try out this release, you can install it with the following commands:

``` bash
$ npm cache clean router
$ npm install pillarjs/router#2.0
```

Owners/collaborators: please do not merge this PR :)

PR #92, opened 2020-04-19
Handle rejected promises from `router.param`

### LINKED ISSUE #91: Promise support for `router.param`
This is a reminder to myself (or anyone else) to implement this and give a smooth upgrade path to users of `express-async-errors`

https://github.com/davidbanham/express-async-errors/issues/28

jonchurch:
Closes #91 

Checks if function passed to `router.param` returns a promise, catches rejected promise. 

Ended up reusing the same approach from handling rejected promises from middleware.

Moved the `isPromise` function used for handling middleware promises into a new `util` folder to share the logic without duplicating it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,11 @@ Maps the specified path parameter `name` to a specialized param-capturing middle
 
 This function positions the middleware in the same stack as `.use`.
 
+The function can optionally return a `Promise` object. If a `Promise` object
+is returned from the function, the router will attach an `onRejected` callback
+using `.then`. If the promise is rejected, `next` will be called with the
+rejected value, or an error if the value is falsy.
+
 Parameter mapping is used to provide pre-conditions to routes
 which use normalized placeholders. For example a _:user_id_ parameter
 could automatically load a user's information from the database without
--- a/index.js
+++ b/index.js
@@ -13,6 +13,7 @@
  */
 
 var flatten = require('array-flatten').flatten
+var isPromise = require('is-promise')
 var Layer = require('./lib/layer')
 var methods = require('methods')
 var mixin = require('utils-merge')
@@ -642,7 +643,12 @@ function processParams (params, layer, called, req, res, done) {
     if (!fn) return param()
 
     try {
-      fn(req, res, paramCallback, paramVal, key.name)
+      var ret = fn(req, res, paramCallback, paramVal, key.name)
+      if (isPromise(ret)) {
+        ret.then(null, function (error) {
+          paramCallback(error || new Error('Rejected promise'))
+        })
+      }
     } catch (e) {
       paramCallback(e)
     }
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,8 @@
 
 This incorporates all changes after 1.3.5 up to 1.3.6.
 
+  * Add support for returned, rejected Promises to `router.param`
+
 2.0.0-beta.1 / 2020-03-29
 =========================
 
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -12,6 +12,7 @@
  * @private
  */
 
+var isPromise = require('is-promise')
 var pathRegexp = require('path-to-regexp')
 
 /**
@@ -185,20 +186,6 @@ function decodeParam (val) {
 
     throw err
   }
-}
-
-/**
- * Returns true if the val is a Promise.
- *
- * @param {*} val
- * @return {boolean}
- * @private
- */
-
-function isPromise (val) {
-  return val &&
-    typeof val === 'object' &&
-    typeof val.then === 'function'
 }
 
 /**
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "repository": "pillarjs/router",
   "dependencies": {
     "array-flatten": "3.0.0",
+    "is-promise": "4.0.0",
     "methods": "~1.1.2",
     "parseurl": "~1.3.3",
     "path-to-regexp": "3.2.0",
--- a/test/param.js
+++ b/test/param.js
@@ -10,6 +10,8 @@ var shouldNotHitHandle = utils.shouldNotHitHandle
 var createServer = utils.createServer
 var request = utils.request
 
+var describePromises = global.Promise ? describe : describe.skip
+
 describe('Router', function () {
   describe('.param(name, fn)', function () {
     it('should reject missing name', function () {
@@ -252,6 +254,48 @@ describe('Router', function () {
       request(server)
         .get('/user/bob')
         .expect(500, /Error: boom/, done)
+    })
+
+    describePromises('promise support', function () {
+      it('should pass rejected promise value', function (done) {
+        var router = new Router()
+        var server = createServer(router)
+
+        router.param('user', function parseUser (req, res, next, user) {
+          return Promise.reject(new Error('boom'))
+        })
+
+        router.get('/user/:user', function (req, res) {
+          res.setHeader('Content-Type', 'text/plain')
+          res.end('get user ' + req.params.id)
+        })
+
+        request(server)
+          .get('/user/bob')
+          .expect(500, /Error: boom/, done)
+      })
+
+      it('should pass rejected promise without value', function (done) {
+        var router = new Router()
+        var server = createServer(router)
+
+        router.use(function createError (req, res, next) {
+          return Promise.reject() // eslint-disable-line prefer-promise-reject-errors
+        })
+
+        router.param('user', function parseUser (req, res, next, user) {
+          return Promise.reject() // eslint-disable-line prefer-promise-reject-errors
+        })
+
+        router.get('/user/:user', function (req, res) {
+          res.setHeader('Content-Type', 'text/plain')
+          res.end('get user ' + req.params.id)
+        })
+
+        request(server)
+          .get('/user/bob')
+          .expect(500, /Error: Rejected promise/, done)
+      })
     })
 
     describe('next("route")', function () {


dougwilson:
Annoyingly more often than we wish, folks end up "reaching in" to our package to require files because... reasons (IDK why). We should probably just drop in this module instead https://github.com/then/is-promise

jonchurch:
You mean so no one can try and require the `isPromise` util we'd be exposing publicly?

jonchurch:
Ughhh my auto linter ran on the readme, fix incoming

dougwilson:
Yea. But also the bonus that ideally they have a test suite to know what is and is not a promise, unlike our coped-in one. A module like that (which has no dependencies) is an easy one to add, as we can depend on a specific version without risk of module hijacking.

jonchurch:
Okay switched to using the `is-promise` package, squashed to cleanup, and then pinned `is-promise` to it's current version.

dougwilson:
I'm not sure why Travis CI failed, as the log is not loading for me, but based on the version numbers, it is likely because the new testes are not contained in a promise test. See https://github.com/pillarjs/router/pull/60/commits/d2bce0c09d424ea89e77cef7743e12f411ed94f7 for an example. Note the `describePromises` helper and wrapping promise tests within it.

README.md
```diff
@@ -115,6 +115,10 @@ Maps the specified path parameter `name` to a specialized param-capturing middle
 
 This function positions the middleware in the same stack as `.use`.
 
+If a `Promise` object is returned from the `param_middleware` function, the router
```
wesleytodd, line 118:
```suggestion
If a `Promise` is returned from the `param_middleware` function, the router
```

README.md
```diff
@@ -115,6 +115,10 @@ Maps the specified path parameter `name` to a specialized param-capturing middle
...
+will attach an `onRejected` callback using `.then`. If the promise is rejected,
```
wesleytodd, line 119:
```suggestion
will attach an `onRejected` callback using `.then(null, onRejected)`. If the promise is rejected,
```

I only think this could use more explicit call out because there is no commonly agreed upon wording I have seen other places.  Having the full explicit call might help make it more clear.

README.md
```diff
@@ -115,6 +115,10 @@ Maps the specified path parameter `name` to a specialized param-capturing middle
...
+`next` will be called with the rejected value, or an error if the value is falsy.
```
wesleytodd, line 120:
```suggestion
`next` will be called with the rejected value as the error, or `new Error...` if the value is falsy.
```

index.js
```diff
@@ -634,7 +635,12 @@ function processParams (params, layer, called, req, res, done) {
     if (!fn) return param()
 
     try {
-      fn(req, res, paramCallback, paramVal, key.name)
+      var ret = fn(req, res, paramCallback, paramVal, key.name)
+      if (isPromise(ret)) {
+        ret.then(null, function (error) {
+          paramCallback(error || new Error('Rejected promise'))
```
wesleytodd, line 641:
```suggestion
          var err  = error || new Error('Rejected promise from param callback')
          if (Error.captureStackTrace) {
            Error.captureStackTrace(fn, err)
          }
          paramCallback(error || new Error('Rejected promise'))
```

I might be wrong, so this would be an awesome test to add, but on a rejected promise I think the error will have a short stack trace.  I *think* this would ensure that the stack has more value when reported.  Again, I think testing this would be great.  If not the stack trace I think a message better than `rejected promise` would be good.

wesleytodd (dismissed):

index.js
dougwilson: This was just copying the current behavior for handling rejections from middlewares, for example. Ideally what is done here is consistent between params and middlewares, so if we do think it should change, we should queue up a separate PR that can land either before or after this one with said changes.

Now, that said, this change is unlikely to do anything, as the first argument will skip the stack frames up to where that function is. So in essence, this call reduces the number of stack frames in the trace, but you seem to be saying you think there is not enough as there is?

Perhaps you can show what the before/after looks like for this change?

jonchurch: I put together a sandbox, which does not test the `router.param` feature, but tests the rejected middleware feature.

https://codesandbox.io/s/github/jonchurch/express-issues/tree/async-stacks

There's three routes to test: `'/'` `'/test'` and `'/middleware'`

Please do take a look.

dougwilson:
So I believe all the comments have been addressed now. Any objections to merging at this point?

wesleytodd:
I wound't consider them blocking, but my comments on the docs are not fully addressed.  So no opposition to merging, but I stand by my statements on the docs changes.

dougwilson:
Apologies Wes, I missed those doc comments when looking over. Yes, the doc suggestions make sense to me to land in the PR. This is exactly why i wanted to ask before doing anything, to make sure nothing got missed in all the comments above :)

jonchurch:
Been spaced these past couple weeks, dunno why I didn't commit the doc changes when they came in hehe. I committed the doc suggestions as is @wesleytodd. 

I haven't done much testing around the stack trace changes suggested. I think it's definitely worth looking into. It pertains to the middleware rejection catching too though, so it's probably best addressed in another PR.

index.js
wesleytodd: Yeah I think my point was that while we are making changes, it might be nice to also improve the stack traces.  In this case, and the other middleware cases, we can provide better debugability with `.captureStackTrace`.  The example from @jonchurch seems to illustrate the issue but typically see this with normal callback code so I would have to test a bit to be sure.

The goal would be having a trace which included where the middleware function was defined, pointing to user code over express code.

dougwilson: Yep, definately @wesleytodd ! I took a look at @jonchurch sandbox, and the error stack trace seems to do exactly that: it shows the user-location and not a bunch of stack within Express. Perhaps I'm just still not understanding what is missing here. Can you provide a similar sandbox showing what you are saying the issue is @wesleytodd ?

For reference, the sandbox shows the error as
```
Error: Where did I occur?
    at wait.then (/sandbox/index.js:19:37)
```

And this is like 19 +/- 1:
```js
app.get('/', (req, res, next) =>
  wait(3).then(() => Promise.reject(new Error('Where did I occur?')))
);
```

dougwilson:
So the doc changes have been committed here, and I know there is still an ongoing thread around the stack traces. Is the stack trace discussion something to open a separate issue or PR about, or do we want to continue it here prior to landing? Mainly looking for confirmation from @wes

mastermatt:
This was brought up in the TC meeting yesterday.

Is the use of `is-promise` (or any Promise sniffer) being used because v2.0 is going to be maintaining Node support <0.12?

dougwilson:
I believe the reason is that we are not _requiring_ everyone re-write their handlers & middlewares as async function or to return a promise, so the code needs to determine if there was a returned value was was a promise or not. I looked through on promises API in the standards and there does not seem to be any built-in method to detect if something is a promise or not, is that right? The closest would be `instanceof Promise`, thought that would not work if users are returning user-land-based promises (like those from bluebird, et. al.).

mastermatt:
My understanding is that the expectation when wanting to support promise and non-promise in the same flow is to put the variable into `Promise.resolve`. But this only works in Node 0.12+.

```js
try {
    var ret = fn(req, res, paramCallback, paramVal, key.name)
    Promise.resolve(ret).then(null, function (error) {
        paramCallback(error || new Error('Rejected promise'))
    })
} catch (e) {
    paramCallback(e)
}
```

dougwilson:
Won't that hurt the performance for the folks who are not using premises, though? We should probably see what the perf cost is going to be between the two methods. We wouldn't want to unnecessarily hurt the perf of folks if not necessary.

mastermatt:
Good point. It will queue a micro-task either way. For `paramCallback` it doesn't matter since the result isn't used, but in the layer it would cause `next` to not be called for an extra tick.

dougwilson:
Right, I would agree. That and the additional work to create the promise object and associated data. Now, what I don't _know_ is how much it actually hurts :) It's certainly more operations, etc. but how much wall clock time it actually adds is to be seen, lol.

alexandernanberg:
Seems like this is ready to be merged into v2, unless I'm missing something?

dougwilson:
You can see above there is a requested change from a reviewer that has not been resolved yet.

index.js
jonchurch: After testing the suggested changes (on Node.js v14) I didn't find a difference between the stack traces.

Instead of testing the changes specifically to `router.param` I tested the equivalent change to `router.handleRequest` which is where the logic in this PR was copied from. 


```md
**As per PR**

/
Error: Where did I occur?
    at /home/box/Projects/express-issues/index.js:19:37

/test
Error: Where did I throw?
    at iThrow (/home/box/Projects/express-issues/anotherModule.js:2:9)
    at /home/box/Projects/express-issues/index.js:9:9

/middleware
Error: Where did I throw?
    at iThrow (/home/box/Projects/express-issues/anotherModule.js:2:9)
    at badMiddleware (/home/box/Projects/express-issues/index.js:23:9)
    at Layer.handleRequest (/home/box/Projects/router/lib/layer.js:103:15)
    at next (/home/box/Projects/router/lib/route.js:142:13)
    at Route.dispatch (/home/box/Projects/router/lib/route.js:107:3)
    at handle (/home/box/Projects/router/index.js:422:11)
    at Layer.handleRequest (/home/box/Projects/router/lib/layer.js:103:15)
    at /home/box/Projects/router/index.js:289:22
    at processParams (/home/box/Projects/router/index.js:569:12)
    at next (/home/box/Projects/router/index.js:283:5)

---------------------------

**With proposed changes:**

/
Error: Where did I occur?
    at /home/box/Projects/express-issues/index.js:19:37

/test
Error: Where did I throw?
    at iThrow (/home/box/Projects/express-issues/anotherModule.js:2:9)
    at /home/box/Projects/express-issues/index.js:9:9

/middleware
Error: Where did I throw?
    at iThrow (/home/box/Projects/express-issues/anotherModule.js:2:9)
    at badMiddleware (/home/box/Projects/express-issues/index.js:23:9)
    at Layer.handleRequest (/home/box/Projects/router/lib/layer.js:102:15)
    at next (/home/box/Projects/router/lib/route.js:142:13)
    at Route.dispatch (/home/box/Projects/router/lib/route.js:107:3)
    at handle (/home/box/Projects/router/index.js:422:11)
    at Layer.handleRequest (/home/box/Projects/router/lib/layer.js:102:15)
    at /home/box/Projects/router/index.js:289:22
    at processParams (/home/box/Projects/router/index.js:569:12)
    at next (/home/box/Projects/router/index.js:283:5)    
```

gilly3:
It looks like the only thing @wesleytodd was still waiting for from the [requested changes](#pullrequestreview-412250725) was an improvement to the captured stack trace for better debuggability.  And then, that improvement seems to have been shown to make no actual difference in the resulting stack traces.

Are we waiting for more proof that the stack traces either can or cannot be improved?  Or is there something else holding this up?

dougwilson:
Hi @gilly3 thanks for the question. I will take a look at the state of this PR.

dougwilson (approved):

PR #93, opened 2020-05-27
feat: removeAllRoutes public method

robertsLando:
This method allows to clear routes stack. As discussed here: https://github.com/expressjs/express/issues/4296

### DIFF
--- a/test/router.js
+++ b/test/router.js
@@ -35,6 +35,21 @@ describe('Router', function () {
     router({}, {}, done)
   })
 
+  describe('.removeAllRoutes()', function () {
+    it('should remove all routes from the stack', function (done) {
+      var router = new Router()
+
+      for (var i = 0; i < 10; i++) {
+        router.get('/thing' + i, helloWorld)
+      }
+
+      router.removeAllRoutes()
+
+      assert.equal(router.stack.length, 0)
+      done()
+    })
+  })
+
   describe('.all(path, fn)', function () {
     it('should be chainable', function () {
       var router = new Router()
--- a/index.js
+++ b/index.js
@@ -436,6 +436,16 @@ Router.prototype.process_params = function process_params(layer, called, req, re
 }
 
 /**
+ * Remove all paths from the current stack
+ *
+ * @public
+ */
+
+Router.prototype.removeAllRoutes = function removeAllRoutes() {
+  this.stack = []
+}
+
+/**
  * Use the given middleware function, with optional path, defaulting to "/".
  *
  * Use (like `.all`) will run for any http METHOD, but it will not add
--- a/index.js
+++ b/index.js
@@ -436,7 +436,9 @@ Router.prototype.process_params = function process_params(layer, called, req, re
 }
 
 /**
- * Remove all paths from the current stack
+ * Removes all paths from the current stack
+ *
+ * Removing routes could be useful when you need to add/remove routes on runtime.
  *
  * @public
  */


robertsLando:
In my case, I have some plugins that a user can add to the running server, those plugins can be customized and can also add apis (routes) to the server. When a plugin is refreshed I need to remove all eisting routes for that plugin and add back the new updated routes.

> would a use expect that calling this would reset the registered parameters

I don't know what you mean as I didn't checked all the source code. By consider my use case do you think that is necessary? Please let me know if I need to add something or feel free to edit my pr.

I added a test, let me know if it's good for you

dougwilson:
Based on your use case description, I'm not sure, but it sounds like your plugin should be returning a new router each time it is "refreshed", which would make this addition not necessary, right? What does this addition add or make possible vs returning a new router fresh for each refresh?

For example, would a user who removes all the routes expect that the router kept the prior strict routing behavior? What if the user wanted to remove all the routes in order to add new routes with a different strict behavior? I guess the same could be asked about the case-sensitivity feature. Those switches are of course aspects of how the routes are interpreted, same as for the params.

For example, say you have a router with some routes like `/user/:id`. You get a reference to that router and want to remove all the routes, so call this new method. You then add new ones like `/obj/:id` and of course call `.param` to define what the `:id` will look like... but it's different than that user one and now conflicts with the previous definition.

So the above scenario may be confusing for how one would use this new method as it's currently (as of the comment) implemented, as there are still aspects of the previous routes still lingering on the router even after calling this method. I of course understand this method is called "removeAllRoutes", but without docs in this PR it's still hard for me to understand what that is expected to do.

Of course, most importantly of all the questions, is the first couple in the comment if you can get to them. Many times having multiple ways to do the same thing leads to user confusion, but also if there is already a way to do this _and_ it is more configurable/capable, it seems adding a partial feature may not be the best long-term route to take, but that's what I'm trying to figure out 😂 .

I think your use-case outline above is leaving a lot to be desired, as I think you are accidentally excluding aspects that is important for us to know here. For example, you say "plugins" but there is no such thing in this module or Express, so I really don't know what that means exactly, which makes following your use-case quite difficult.

test/router.js
```diff
@@ -35,6 +35,21 @@ describe('Router', function () {
...
+    it('should remove all routes from the stack', function (done) {
+      var router = new Router()
+
+      for (var i = 0; i < 10; i++) {
+        router.get('/thing' + i, helloWorld)
+      }
+
+      router.removeAllRoutes()
+
+      assert.equal(router.stack.length, 0)
```
dougwilson, line 48: Tests should only be asserting against the public-visible behavior/API. I assume that by adding this method that you are saying that `router.stack` is private and not public. This test using `.stack` to make an assertion would seemingly re-enforce the idea that it is public (as changing the name of the property would break this test -- the indicator it's not private). If it is public, then I'm missing the purpose of this function, as `router.stack = []` would already be clearly possible for a user to do, is that right?

jonchurch:
Can you provide us a sample app that would benefit from this feature? Maybe a comparison w/ how things are now, and how you would like them to work with the inclusion of this feature.

This is an interesting feature, but it also feels a lot like state management which I'm wary of. The route stack is of course already a type of application state, and Express is very unopinionated so if you want to manage routes as state that's up to the user. But something feels off about providing an API that encourages users to treat the route stack like state that can be wiped with a single function call.

test/router.js
robertsLando: Yes I could do that already, but having it in the router as a function makes it more "safe" against updates that could break that

dougwilson:
> This allows me to keep using the same router instance without creating another one or use the trick you mentioned here

So I think your view of it as "trick" is not correct -- it it no trick at all, just code to implement what you want. There is nothing wrong with it.

> > I think you are accidentally excluding aspects that is important for us to know here.

> That  is sure as I don't know ROuter as you and I didn't checked all the source code 😄

Oh, I think I may have miss-spoke here. I'm referring to excluding aspects of your use-case us, and important for us to know about your use-case. We understand you don't know the router code, and that is what we want to help with, but are still unclear on your use-case.

I think what @jonchurch said in https://github.com/pillarjs/router/pull/93#issuecomment-634757459 would be helpful for sure 👍

robertsLando:
> This is an interesting feature, but it also feels a lot like state management which I'm wary of. The route stack is of course already a type of application state, and Express is very unopinionated so if you want to manage routes as state that's up to the user. But something feels off about providing an API that encourages users to treat the route stack like state that can be wiped with a single function call.

Keep it like a proposal guys, I saw some questions around stackoverflow and plugins like https://www.npmjs.com/package/express-remove-route that would like someting to manage routes. In my case this is enought but could open a new way of features around routes management (add/remove specific ones for example)

wesleytodd:
I think it sounds like this was agreed to be better done with a new router instance. So I am going to close this. If you want to open a new issue to discuss alternatives please do. One alternative I think worth discussing is showing folks how to easily use a new router to achieve this behavior.

PR #100, opened 2021-04-27
Add typescript definition files

Rauno56:
- Rebased #76,
- Brought typescript up to date,
- Added all HTTP methods, as requested in the original PR,
- Removed usages of `any` where possible,
- Fixed some bugs in the definitions.


index.d.ts
```diff
@@ -0,0 +1,202 @@
...
+     */
+    all: IRouterMatcher<this>;
+
+    use: IRouterHandler<this> & IRouterMatcher<this>;
+
+    handle: (req: http.IncomingMessage, res: http.ServerResponse, cb: Callback) => void;
+
+    route(prefix: PathParams): IRoute;
+    // Stack of configured routes
+    stack: Layer[];
```
dougwilson, line 108: As a general question, are the typings supposed to include all properties and methods including private ones, so should this be removed since it is not a public property?

index.d.ts
```diff
@@ -0,0 +1,202 @@
...
+    link: IRouterMatcher<this>;
+    lock: IRouterMatcher<this>;
+    "m-search": IRouterMatcher<this>;
+    merge: IRouterMatcher<this>;
+    mkactivity: IRouterMatcher<this>;
+    mkcalendar: IRouterMatcher<this>;
+    mkcol: IRouterMatcher<this>;
+    move: IRouterMatcher<this>;
+    notify: IRouterMatcher<this>;
+    pri: IRouterMatcher<this>;
```
dougwilson, line 134: Is this actually a method on there?
```
$ node -pe 'require(".")().pri'
undefined
```

index.d.ts
```diff
@@ -0,0 +1,202 @@
...
+     *          }
+     *        });
+     *      });
+     */
+    param(name: string, handler: RequestParamHandler): this;
+
+    /**
+     * Alternatively, you can pass only a callback, in which case you have the opportunity to alter the app.param()
+     *
+     * @deprecated since version 4.11
```
dougwilson, line 92: This module is still in the 1.x version range, so the version here doesn't seem right.

index.d.ts
```diff
@@ -0,0 +1,202 @@
...
+  interface Layer {
+    name: string,
+    handle: RequestHandler | ErrorRequestHandler,
+    handle_request: RequestHandler,
+    handle_error: ErrorRequestHandler,
+    match: (path: string) => boolean,
+  }
+
+  interface IncomingRequest extends http.IncomingMessage {
+    url: string,
```
dougwilson, line 22: Isn't the url property already a member of http.IncomingMessage ? (please forgive me on some of these typescript questions -- they may be obvious to someone who knows typescript).

index.d.ts
Rauno56: I took that from `require('http').METHODS`. `v14.16.1` has it at least.

index.d.ts
Rauno56: Yes... that must've come from #76. Removed.

index.d.ts
Rauno56: Nope. That doesn't have to be there. Removed.

I'm not an expert on TS either - it would be nice to have someone, who fits the description, go over it though.

index.d.ts
Rauno56: You are right. API documented private in the code should be defined private in TS as well. I didn't pay much attention to that since

- "it's JS - accessible anyways!" and
- I've actually needed to access those myself.

I'll go over it and add access modifiers where appropriate next week.

index.d.ts
dougwilson: Ah, gotcha. I was using Node.js 16.1. So that does bring up a good point: I guess this should be all the methods _possible_ in the various Node.js versions? I wonder what other methods got removed; probably should go through all the versions this module supports and union them together.

index.d.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  type RequestParamHandler = (req: IncomingRequest, res: http.ServerResponse, next: NextFunction, value: string, name: string) => void;
+
+  type RouteHandler = (req: RoutedRequest, res: http.ServerResponse, next: NextFunction) => void;
+  type RequestHandler = (req: IncomingRequest, res: http.ServerResponse, next: NextFunction) => void;
+
+  type NextFunction = (err?: Error | "route" | "router") => void;
+  type Callback = (err?: Error) => void;
+
+  type ErrorRequestHandler = (err: Error, req: IncomingRequest, res: http.ServerResponse, next: NextFunction) => void;
```
dougwilson, line 42: We may not want the `err` to be typed as just a `Error` object, as in the 1.x of this router, there is no enforcement on that; using a middleware you didn't write could have a string or plain object here, which I've seen a lot. We wouldn't want someone's TS code to think they don't need some kind of guard on the `err` value.

Rauno56:
Rebased and reimplemented the CI changes on GHA.

index.d.ts
Rauno56: I asked around in TS discord and they didn't recall a better way to handle it than what you proposed - adding collecting all of them from all versions.

terry-au:
Is there anything blocking this PR from being merged into master?

index.d.ts
Rauno56: What's more appropriate type for that then? `any`?

index.d.ts
Rauno56: @dougwilson, I looked into it a bit, but don't know a good way to implement it. I'd leave it as it is for now.

Rauno56:
Except for https://github.com/pillarjs/router/pull/100#discussion_r660599459, I would not block it.

index.d.ts
dougwilson: I assume the way to implement is to not have it in here at all. My understanding (which is weak, please correct me if wrong) is that the TSD is just the public API, which is why there is no mechanism to mark something as private.

index.d.ts
dougwilson: That is probably fine. It does have to be truthy, though, so not sure if that helps at all.

index.d.ts
```diff
@@ -0,0 +1,193 @@
...
+    },
+  }
+
+  interface RoutedRequest extends IncomingRequest {
+    baseUrl: string,
+    next?: NextFunction,
+    route?: IRoute
+  }
+
+  type RequestParamHandler = (req: IncomingRequest, res: http.ServerResponse, next: NextFunction, value: string, name: string) => void;
```
dougwilson, line 34: I tried to figure out where it went wrong in the TS, and I think this line is the culprit: inside of the `router.param` callback, there is `req.params`, `req.route`, and all the other things available. I think that this should be the "routed request" type here.

index.d.ts
Rauno56: Commented out `handle()` and `stack` - left Layer type because it's useful for documenting the internal API, which will be inaccessible nevertheless.

index.d.ts
Rauno56: Changed it to `any`.

index.d.ts
Rauno56: Changed that to `RoutedRequest`.

PR #102, opened 2021-11-15
deps: path-to-regexp@6.2.0

nfantone:
* Bump `path-to-regexp` to `6.2.0`.
* Add tests for new features introduced since `path-to-regexp@3.2.0`.
* Fix minor typos in test descriptions and some `.md` files.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,11 @@
 This incorporates all changes after 1.3.5 up to 1.3.7.
 
   * Add support for returned, rejected Promises to `router.param`
+  * deps: path-to-regexp@6.2.0
+    - Support for **named capturing groups** in paths using `RegExp`, i.e.: `/\/(?<group>.+)/` would result in `req.params.group` being populated.
+    - Custom **prefix and suffix groups** using `{}`, e.g.: `/:entity{-:action}?` would match `/user` and `/user-delete`.
+    - Unbalanced patterns would now produce an error: `/test(foo` previously would worked, now it expects either `(` to be closed or the character to be escaped for the previous behavior, e.g. `/test\\(foo`.
+    - Just like with parentheses since `2.0.0-beta.1`, bracket literals now require escaping, i.e.: `/user{:id}` no longer matches `/user{42}` as before unless written as `/user\\{:id\\}`.
 
 2.0.0-beta.1 / 2020-03-29
 =========================
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,8 +7,8 @@ This incorporates all changes after 1.3.5 up to 1.3.7.
   * deps: path-to-regexp@6.2.0
     - Support for **named capturing groups** in paths using `RegExp`, i.e.: `/\/(?<group>.+)/` would result in `req.params.group` being populated.
     - Custom **prefix and suffix groups** using `{}`, e.g.: `/:entity{-:action}?` would match `/user` and `/user-delete`.
-    - Unbalanced patterns would now produce an error: `/test(foo` previously would worked, now it expects either `(` to be closed or the character to be escaped for the previous behavior, e.g. `/test\\(foo`.
-    - Just like with parentheses since `2.0.0-beta.1`, bracket literals now require escaping, i.e.: `/user{:id}` no longer matches `/user{42}` as before unless written as `/user\\{:id\\}`.
+    - Unbalanced patterns now produce an error: `/test(foo` previously worked, but now it expects `(` to be escaped for the previous behavior, e.g. `/test\\(foo`.
+    - Just like with parentheses since `2.0.0-beta.1`, bracket literals now require escaping, i.e.: `/user{:id}` no longer matches `/user{42}` as before unless written as  `/user\\{:id\\}`.
 
 2.0.0-beta.1 / 2020-03-29
 =========================
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "is-promise": "4.0.0",
     "methods": "~1.1.2",
     "parseurl": "~1.3.3",
-    "path-to-regexp": "3.2.0",
+    "path-to-regexp": "6.2.0",
     "setprototypeof": "1.2.0",
     "utils-merge": "1.0.1"
   },
--- a/test/route.js
+++ b/test/route.js
@@ -687,7 +687,7 @@ describe('Router', function () {
             .expect(200, { 0: 's', user: 'tj', op: 'edit' }, cb)
         })
 
-        it('should work inside literal paranthesis', function (done) {
+        it('should work inside literal parenthesis', function (done) {
           var router = new Router()
           var route = router.route('/:user\\(:op\\)')
           var server = createServer(router)
@@ -697,6 +697,27 @@ describe('Router', function () {
           request(server)
             .get('/tj(edit)')
             .expect(200, { user: 'tj', op: 'edit' }, done)
+        })
+
+        it('should allow matching literal parenthesis within a group', function (done) {
+          var cb = after(3, done)
+          var router = new Router()
+          var route = router.route('/:user([a-z\\(\\)]+)')
+          var server = createServer(router)
+
+          route.all(sendParams)
+
+          request(server)
+            .get('/1234')
+            .expect(404, cb)
+
+          request(server)
+            .get('/foo')
+            .expect(200, { user: 'foo' }, cb)
+
+          request(server)
+            .get('/(foo)')
+            .expect(200, { user: '(foo)' }, cb)
         })
 
         it('should work within arrays', function (done) {
--- a/test/route.js
+++ b/test/route.js
@@ -14,6 +14,8 @@ var shouldHitHandle = utils.shouldHitHandle
 var shouldNotHaveBody = utils.shouldNotHaveBody
 var shouldNotHitHandle = utils.shouldNotHitHandle
 
+// Named capturing groups are available from Node `10.0.0` and up
+var describeCaptureGroups = runningOnNodeMajorVersionGreaterOrEqualThan(10) ? describe : describe.skip
 var describePromises = global.Promise ? describe : describe.skip
 
 describe('Router', function () {
@@ -1019,60 +1021,57 @@ describe('Router', function () {
         })
       })
 
-      // Named capturing groups are available from Node `10.0.0` and up
-      if (runningOnNodeMajorVersionGreaterOrEqualThan(10)) {
-        describe('using "(?<name>)"', function () {
-          it('should allow defining capturing groups using regexps', function (done) {
-            var cb = after(3, done)
-            var router = new Router()
-            var route = router.route(/\/(?<name>.+)/)
-            var server = createServer(router)
+      describeCaptureGroups('using "(?<name>)"', function () {
+        it('should allow defining capturing groups using regexps', function (done) {
+          var cb = after(3, done)
+          var router = new Router()
+          var route = router.route(/\/(?<name>.+)/)
+          var server = createServer(router)
 
-            route.all(sendParams)
+          route.all(sendParams)
 
-            request(server)
-              .get('/')
-              .expect(404, cb)
+          request(server)
+            .get('/')
+            .expect(404, cb)
 
-            request(server)
-              .get('/foo/bar')
-              .expect(200, { name: 'foo/bar' }, cb)
+          request(server)
+            .get('/foo/bar')
+            .expect(200, { name: 'foo/bar' }, cb)
 
-            request(server)
-              .get('/foo')
-              .expect(200, { name: 'foo' }, cb)
-          })
-
-          it('should work with multiple named groups in the same segment', function (done) {
-            var cb = after(5, done)
-            var router = new Router()
-            var route = router.route(/\/(?<filename>.*).(?<ext>html|pdf|json)/)
-            var server = createServer(router)
-
-            route.all(sendParams)
-
-            request(server)
-              .get('/foo')
-              .expect(404, cb)
-
-            request(server)
-              .get('/foo.xml')
-              .expect(404, cb)
-
-            request(server)
-              .get('/foo.html')
-              .expect(200, { filename: 'foo', ext: 'html' }, cb)
-
-            request(server)
-              .get('/bar.pdf')
-              .expect(200, { filename: 'bar', ext: 'pdf' }, cb)
-
-            request(server)
-              .get('/baz.json')
-              .expect(200, { filename: 'baz', ext: 'json' }, cb)
-          })
+          request(server)
+            .get('/foo')
+            .expect(200, { name: 'foo' }, cb)
         })
-      }
+
+        it('should work with multiple named groups in the same segment', function (done) {
+          var cb = after(5, done)
+          var router = new Router()
+          var route = router.route(/\/(?<filename>.*).(?<ext>html|pdf|json)/)
+          var server = createServer(router)
+
+          route.all(sendParams)
+
+          request(server)
+            .get('/foo')
+            .expect(404, cb)
+
+          request(server)
+            .get('/foo.xml')
+            .expect(404, cb)
+
+          request(server)
+            .get('/foo.html')
+            .expect(200, { filename: 'foo', ext: 'html' }, cb)
+
+          request(server)
+            .get('/bar.pdf')
+            .expect(200, { filename: 'bar', ext: 'pdf' }, cb)
+
+          request(server)
+            .get('/baz.json')
+            .expect(200, { filename: 'baz', ext: 'json' }, cb)
+        })
+      })
     })
   })
 })
--- a/test/route.js
+++ b/test/route.js
@@ -14,8 +14,7 @@ var shouldHitHandle = utils.shouldHitHandle
 var shouldNotHaveBody = utils.shouldNotHaveBody
 var shouldNotHitHandle = utils.shouldNotHitHandle
 
-// Named capturing groups are available from Node `10.0.0` and up
-var describeCaptureGroups = runningOnNodeMajorVersionGreaterOrEqualThan(10) ? describe : describe.skip
+var describeNamedCaptureGroups = supportsRegExp('(?<foo>.)') ? describe : describe.skip
 var describePromises = global.Promise ? describe : describe.skip
 
 describe('Router', function () {
@@ -1021,7 +1020,7 @@ describe('Router', function () {
         })
       })
 
-      describeCaptureGroups('using "(?<name>)"', function () {
+      describeNamedCaptureGroups('using "(?<name>)"', function () {
         it('should allow defining capturing groups using regexps', function (done) {
           var cb = after(3, done)
           var router = new Router()
@@ -1095,6 +1094,10 @@ function sendParams (req, res) {
   res.end(JSON.stringify(req.params))
 }
 
-function runningOnNodeMajorVersionGreaterOrEqualThan (version) {
-  return Number(process.versions.node.split('.')[0]) >= Number(version)
+function supportsRegExp (source) {
+  try {
+    return RegExp(source).source !== ''
+  } catch (e) {
+    return false
+  }
 }
--- a/test/route.js
+++ b/test/route.js
@@ -860,6 +860,54 @@ describe('Router', function () {
         })
       })
 
+      describe('using "{<prefix>:name}"', function () {
+        it('should allow defining custom prefixes', function (done) {
+          var cb = after(3, done)
+          var router = new Router()
+          var route = router.route('/{$:foo}{$:bar}?')
+          var server = createServer(router)
+
+          route.all(sendParams)
+
+          request(server)
+            .get('/bar')
+            .expect(404, cb)
+
+          request(server)
+            .get('/$bar')
+            .expect(200, { foo: 'bar' }, cb)
+
+          request(server)
+            .get('/$bar$fizz')
+            .expect(200, { foo: 'bar', bar: 'fizz' }, cb)
+        })
+
+        it('should work in any segment', function (done) {
+          var cb = after(4, done)
+          var router = new Router()
+          var route = router.route('/user/:foo?{-:bar}?{-:baz}?')
+          var server = createServer(router)
+
+          route.all(sendParams)
+
+          request(server)
+            .get('/user')
+            .expect(200, cb)
+
+          request(server)
+            .get('/user/delete')
+            .expect(200, { foo: 'delete' }, cb)
+
+          request(server)
+            .get('/user/delete-fizz')
+            .expect(200, { foo: 'delete', bar: 'fizz' }, cb)
+
+          request(server)
+            .get('/user/delete-fizz-buzz')
+            .expect(200, { foo: 'delete', bar: 'fizz', baz: 'buzz' }, cb)
+        })
+      })
+
       describe('using "(regexp)"', function () {
         it('should add capture group using regexp', function (done) {
           var cb = after(2, done)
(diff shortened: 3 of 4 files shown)


dougwilson:
What are the changes between 3.x and 6.x? We need to document and add tests for whatever new features that were added to route paths, if any. The test suite is alsl unfortunately not comprehensive, so no failure does not mean there isn't some breaking change we need to document, and perhaps test to add. Getting a list of what the changes were in the dependency would help determine that.

For example, the "Custom Prefix and Suffix" system I believe was added in 6.0 and of course we have no tests or any docs around that. I'm sure there are other changes, but I know that is at least one.

nfantone:
@dougwilson I 100% understand where you're coming from and I'm all for doing this at some point -  but may I ask you to elaborate on why you think we should do it now? `path-to-regexp`is an internal implementation detail. The API surface of `router` hasn't changed; we haven't added any new features to this module. Everything that we _want_ to support and is documented still works. I'm not sure I see a need for what you're requesting.

dougwilson:
The path-to-regexp module is not an internal implementation detail; it is actually the main component of this module's (and Express's) public API: the route strings. Are you suggesting that Express should not document what kind of things you can put in the argument for something like `app.get`?

dougwilson:
Here is a breaking change I just found with this upgrade that breaks the `router` public API and should be documented:
The path `'/:foo([a-z()]+)'` would construct a path that captures a URL that contains the characters `a-z` and `(`, and `)`, but with this upgrade, that path now throws with `TypeError: Capturing groups are not allowed at 10`.

I'm sure I'm just scratching the surface of the breaking changes in how the paths work with these two major version bumps, as I just stopped once I saw the first breaking change in path-to-regexp dependency. As noted, the test suite in here is not very great, especially when a PR like this adds no tests in order to detect future breaking changes :)

nfantone:
> The `path-to-regexp` module is not an internal implementation detail; it is actually the main component of this module's (and Express's) public API.

Gotcha. I disagree, though. From my POV, it most definitely is an implementation detail. How much `router` relies on it is also an implementation details. Users installing `router` don't know nothing about `path-to-regexp`, the same way they know nothing about `methods` or `array-flatten`.

> Are you suggesting that Express should not document what kind of things you can put in the argument for something like `app.get`,

No. I'm suggesting that it is _already_ documented and supported and there's no further need to more details.

nfantone:
> Here is a breaking change I just found with this upgrade that breaks the router public API and should be documented

Agreed, then. If it breaks the existing public API, it should be tested and documented. Good catch.

dougwilson:
Gotcha. So I get it, they should not know path-to-regexp exists today, apologies there. I think we are on the same page. So when a user sees the following path `app.get('/user{-:id}', fn)` where do they go to find out what that is supposed to do, exactly? And also, does the `req.params` this module is constructing for that new path-to-regexp feature even make sense?

nfantone:
Perhaps the problem with my focus is that I may be relying a bit too much on the test suite here. I saw all `703` tests passing and I assumed everything was working as expected. Having said that, my answers to your questions would be:

> So when a user sees the following path `app.get('/user{-:id}', fn)` where do they go to find out what that is supposed to do, exactly? 

It depends. Does that exist today, i.e.: is that supported with latest Express 5 beta release? If it is, there should already be docs for that (if there aren't, we should add them). If it's not, then there wouldn't be a specific place as it is simply not supposed to work with this release.

I guess what I'm trying to convey here is that `express` defines the features that `path-to-regexp` should provide. Not the other way around.

dougwilson:
> Perhaps the problem with my focus is that I may be relying a bit too much on the test suite here. I saw all 703 tests passing and I assumed everything was working as expected.

Right, as I said, the test suite is not a good indicator. It is very lacking. I have over time added a lot more tests, but there is still more to add. It doesn't always keep up, either, like this PR doesn't add any tests for all the new path matching added in the version upgrade.

> It depends. Does that exist today, i.e.: is that supported with latest Express 5 beta release? If it is, there should already be docs for that (if there aren't, we should add them). If it's not, then there wouldn't be a specific place as it is simply not supposed to work with this release.

It doesn't exist in the current Express 5, as that is based off path-to-regexp 0.1.x. Your PR here is what is adding that feature, so it would not exist in any current documentation, which is why I'm saying should it be added somewhere as part of this upgrade which adds this feature.

> I guess what I'm trying to convey here is that express defines the features that path-to-regexp should provide. Not the other way around.

That is an interesting take. I don't believe the path-to-regexp project is run like that, so maybe we need to bring that up with that project's maintainers.

nfantone:
>  Your PR here is what is adding that feature, so it would not exist in any current documentation,

My point is that it's really _not_ adding it. `path-to-regexp` has that feature - `router` doesn't. It may in the near future, but currently it doesn't.

> OR there should be logic added here that filters out the features from path-to-regexp to just the ones that we want to expose.

Right. This is what I was trying to say. It's completely fine to let `path-to-regexp` run its course - but adapting `express` to match its API feels backwards.

Anyway, regardless, I've added a bunch of new tests for all new features I could find on `path-to-regexp@6.2.0`. And a few that cover the breaking change you described before.

dougwilson:
> My point is that it's really not adding it. path-to-regexp has that feature - router doesn't. It may in the near future, but currently it doesn't.

But from a user's perspective who is using `router`, they wouldn't be able to make a route with those features, but then they upgraded `router` and now they can. As far as the user is concerned, `router` added that feature (because path-to-regexp is an internal implementation detail, not something the user would think about).

> Right. This is what I was trying to say. It's completely fine to let path-to-regexp run its course - but adapting express to match its API feels backwards.

Gotcha. So what should we do instead?

HISTORY.md
```diff
@@ -1,171 +1,171 @@
-2.0.0-beta.1 / 2020-03-29
+`2.0.0-beta.1` / 2020-03-29
```
dougwilson, line 1: These headers shouldn't have the backticks; it deviates from all the other history file formats and breaks our tooling.

test/route.js
```diff
@@ -681,6 +682,27 @@ describe('Router', function () {
             .expect(200, { user: 'tj', op: 'edit' }, done)
         })
 
+        it('should allow matching literal parenthesis within a group', function (done) {
+          var cb = after(3, done)
+          var router = new Router()
+          var route = router.route('/:user([a-z\\(\\)]+)')
```
dougwilson, line 688: The PR should include a note in the history entry that these (and whatever other chars) now need to be escaped so folks know.

HISTORY.md
nfantone: Sorry - couldn't resist the temptation. Reverted.

test/route.js
nfantone: Shouldn't that be created when releasing?

dougwilson: No, because things get lost. The history file is updated with every commit as part of the commit itself.

nfantone: Ok. Which heading should I use?

dougwilson: It doesn't matter all too much, as long as the release tooling (under `scripts/`) catches it. I would probably just do `2.x` as that is what the other commits in the 2.0 branch used.

nfantone: Done. Added everything I could find to `HISTORY.md`.

nfantone:
@dougwilson I see you've added test pipelines for very ancient runtimes. This is likely to make tests fail due to the lack of support for widely used language features, such as destructuring, in third-party dependencies.

dougwilson:
I didn't add any new pipelines; this module has always ran tests against all the Node.js versions listed in the `engines` property in `package.json`.

nfantone:
Weird. No pipelines were run in any of my previous commits. It was only after I rebased your latest changes from `1.3.6` that they were triggered.

Also, the `"engines"` property only lists `"node"` - but GH Actions are ran against `io.js` as well.

dougwilson:
> Also, the "engines" property only lists "node" - but GH Actions are ran against io.js as well.

The io.js platform was ths fork of Node.js that then became Node.js again with version 4 (Node.js 4 is io.js 4). It used the same field in engines.

nfantone:
> The project was out of OSS credits in Travis CI at the time.

👍🏼 

>  It used the same field in engines.

Ok. I have oftentimes seen it as `"iojs"`. Which really makes sense since their versioning has diverged.

dougwilson:
> Ok. I have oftentimes seen it as "iojs". Which really makes sense since their versioning has diverged.

I'm not aware of any tooling that used a separate field. Remember, io.js _is_ Node.js now :) https://nodejs.org/en/blog/release/v4.0.0/

nfantone:
> I'm not aware of any tooling that used a separate field

I would know of a bunch. Just search "engines iojs" on all of Github and you'll see it popping out quite a few times.

> If upgrading path-to-regexp no longer works on the Node.js versioned listed in the engines field, then we need to evaluate the impact of bumping the engines requirement in this module and it's consumers (mainly Express).

I couldn't find any tight info on this. `path-to-regexp` doesn't have an `"engines"` field and their changelog/releases don't specify anything in relation to the Node version they support (or, at least, I couldn't dig it out easily).

IMHO, in this day and age, it makes little sense to keep supporting runtimes below LTS. The feature gap is no longer what it used to be and I like to think that we won't be having an angry `io.js` mob shaking fists at us anytime soon.

dougwilson:
> I would know of a bunch. Just search "engines iojs" on all of Github and you'll see it popping out quite a few times.

Can you be more specific? Which ones? I mean, this module has only ever supported being installed by npm and npm never used iojs field when installing on io.js (I even put in a feature request for npm at the time and it was rejected).

> IMHO, in this day and age, it makes little sense to keep supporting runtimes below LTS.

I don't disagree. If this is a blocker for this PR, definitely open up a new issue to track this and you can link yo this PR as support for the reasoning 👍 . We want to make sure that users have a chance to see an upcoming change link this and have a forum to be able to voice their opinion on it. This is our process for dropping things like Node.js versions.

nfantone:
Understood. On which repository should I open it?

dougwilson:
The repo you are desiring to make the change on.

nfantone:
RIght - but this would mean, as you pointed out, changing the min Node version for `express` as a package. So it would impact multiple repositories at once. Also, it'd probably have more visibility to the community if we post such as thing on https://github.com/expressjs/express. Objections?

dougwilson:
There is already an open issue in the express repo, so you can just add to that. But unless express is the only possible repo that consumes this one, there should probably be one here too. I know you believe no one will care, but when you manage this many repos, you come to find out they do. And they of course come out immediately mad at you after you made the release. The purpose of this is mainly so we can point where they had an opportunity to raise their opinion before the change was made and reduces our burden managing these issues. I would expect them to remain upset if we pointed them to a completely different repo, though.

Also, apologies but I have to head off to work, so I won't be around for at least 10 hours.

nfantone:
@dougwilson Here's the relevant issue: https://github.com/pillarjs/router/issues/103.

dougwilson:
> Any reasons why you'd want me to open a new issue asking the same?

I said not to open a new issue to ask that, as there was already one, as you saw :)

nfantone:
You also said to open an issue where you :are desiring to make the change on" :). So I did.

dougwilson:
> You also said to open an issue where you "are desiring to make the change on" :). So I did.

Right. So where have the users of router (an independent project from express) voiced their opinion?

dougwilson:
I looked into the failure and it looks like the only issue is just with the `semver` dependency that was added to the `devDependencies` of this module requires Node.js >= 10. I added the older `semver` to the older Node.js versions so they can use the relevant version of that devDependency and tests are now passing across all versions Node.js 0.10+

HISTORY.md
```diff
@@ -3,35 +3,41 @@
 
 This incorporates all changes after 1.3.5 up to 1.3.6.
 
+* deps: `path-to-regexp@6.2.0` (See [release changelog](https://github.com/pillarjs/path-to-regexp/releases/tag/v6.2.0))
+  - Support for **named capturing groups** in paths using `RegExp`, i.e.: `/\/(?<group>.+)/` would result in `req.params.group` being populated.
+  - Custom **prefix and suffix groups** using `{}`, e.g.: `/:entity{-:action}?` would match `/user` and `/user-delete`.
+  - **BREAKING**: Parenthesis and bracket literals now require escaping, i.e.: `/user(:id)` or `/user{:id}` no longer match `/user(42)`, `/user{42}` as before unless written as `/user\\(:id\\)` or `/user\\{:id\\}`.
```
dougwilson, line 9: I didn't test the `{}` one, since that was added in path-to-regexp 6, but I tested the `/user(:id)` one and `/user(:id)` doesn't match `/user(42)` in 2.0.0-beta.1 either, and needs the escapes. Was this change just for `{}` and was the `()` a typo? I cannot find this noted in `path-to-regexp` changelog, either.

dougwilson: I went ahead and tested `/user{:id}` in 2.0.0-beta.1 as well and it doesn't work with or without `\\` chars. Here is the test I was using to try it out:
```js
it('should match single path segment', function (done) {
  var router = new Router()
  var route = router.route('/user(:id)')

  route.all(sendParams)

  request(server)
    .get('/user(1)')
    .expect(200, '{"id":"1"}', done)
})
```
It returns 404 for both 2.0.0-beta.1 release and on this PR. I'll add these as tests to the test suite as well, as they are good tests to have 👍

jimmywarting:
I don't know if you have heard about this new [URLPattern](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern)
It's pretty sweet if you ask me, would be cool if express used it instead

there is a polyfill floating around in npms user land

HISTORY.md
nfantone: I tested this using `path-to-regexp@6.2.0` itself, if I remember correctly. Yeah, without escaping brackets you'd get a `404`.

```js
var match = require("path-to-regexp").match

match('/user/\\(:id\\)')('/user/(42)')
// {path: "/user/(42)", index: 0, params: {id: "42"}}

m('/user/\\{:id\\}')('/user/{42}')
// {path: "/user/{42}", index: 0, params: {id: "42"}}
```

dougwilson: Right, I understand that. But I guess I wasn't clear in my point: I am saying that I don't think that _changed_ between 2.0.0-beta.1 and this PR, so it doesn't seem like it should be in our History as a breaking change, since nothing broke. For example, using the path-to-regexp that is in 2.0.0-beta.1 directly, as you tested directly with the one in this PR, here are the results on your tests:
```
$ node
Welcome to Node.js v16.7.0.
Type ".help" for more information.
> require("path-to-regexp/package").version
'3.2.0'
> match=require("path-to-regexp").match
[Function: match]
> match('/user/(:id)')('/user/(42)')
false
> match('/user/\\(:id\\)')('/user/(42)')
{ path: '/user/(42)', index: 0, params: { id: '42' } }
> match('/user/{:id}')('/user/{42}')
{ path: '/user/{42}', index: 0, params: { id: '42' } }
> match('/user/\\{:id\\}')('/user/{42}')
{ path: '/user/{42}', index: 0, params: { id: '42' } }
```

You can see that yes, it changes the `{:id}` one, but as I was saying, it doesn't seem to change the parenthesis one at all, right? So when it says "`/user(:id)` [...] no longer match `/user(42)`" in this update, how is that breaking, as it didn't work in the previous version, either.

nfantone: Ah, gotcha 👍🏼. Sorry, misunderstood what you were trying to say. Yeah, you're right. Maybe we could move that item about escaping brackets to the previous release changelog? I don't see it mentioned anywhere else and I think it should be.

dougwilson: Yes, if indeed it does work in router 1.x and then didn't work in router 2.0.0-beta.1 then it was a miss for that changelog and we should add it there. I haven't confirmed it as I am on mobile now, but if confirmed, one of us can update it.

nfantone: Looks like it did work in `path-to-regexp@0.1.7`.

```js
var pathToRegexp = require('path-to-regexp@0.1.7')
'/user/(42)'.match(pathToRegexp('/user/(:id)'))

// ['/user/(42)', '(42)']
```

nfantone:
@dougwilson Do you think we could merge this? Or do you feel like there's more work to be done here? I would really like to move forward with `express` v5 tasks.

dougwilson:
Hi @nfantone I am away for this week for a US holiday and will be back again working on open source after that. I believe there is still an on-going conversion from last I interacted with this PR, including you double checked my thought that part of the history in this PR was incorrect and needed to be updated. We also identified there were no tests around the feature mentioned in the change log, which we could add, though they could just be in their own pr. I know you aksed another question above, but I left already, so I didn't have a chance to evaluate it. I will get back after the holiday 👍

nfantone:
@dougwilson Ping?

dougwilson:
Hi @nfantone thank you for the ping. I just got back today and am currently triaging a security vulnerability report that was given over the holiday. I just need to look in to that ASAP and determine what needs to be done and than I will get back to here 👍 , especially as I'm going to be away again this weekend and the beginning of next week. It does seem this is the time of year (along with March-ish) that a lot of security reports seem to come in for whatever reason.

dougwilson:
In the meantime, please feel free to push the changes discussed above, as we'd need those regardless.

nfantone:
@dougwilson Welcome back. Hope you had a good time.

>  including you double checked my thought that part of the history in this PR was incorrect and needed to be updated

I believe I answered that already in the thread above. Let me know if I'm mistaken.

> We also identified there were no tests around the feature mentioned in the change log

Which feature is that again? Think we've covered them all. Which one did we miss?

package.json
```diff
@@ -29,6 +29,7 @@
     "mocha": "9.1.3",
     "nyc": "15.1.0",
     "safe-buffer": "5.2.1",
+    "semver": "7.3.5",
```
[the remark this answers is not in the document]
dougwilson: For this, as it works with the older version, it doesn't matter to me, which ever you like is fine 👍

dougwilson:
So the changes are that the inaccurate changelog in that discussion above needs to be fixed in this PR, as you listed a breaking change under this PR that didn't change at all in this PR. The tests are for that same feature. https://github.com/pillarjs/router/pull/102#discussion_r751635292

HISTORY.md
dougwilson: Ok, cool. So then the entry does not belong in this change set, as it's just a missing one from 2.0.0-beta.1. It should at least be removed from this changeset. Ideally we should add it over retroactively to the 2.0.0-beta.1 changeset, but you don't need to do that in this PR if you don't want to.

dougwilson:
Oh, and for the tests, sorry, we don't need to add tests for new features that were added before this specific upgrade, as those would just be "misses" from previous PRs and shouldn't hold this one up, but of course you're welcome to do so if you like. The parenthesis stuff was a change from a previous PR; the `{}` matching stuff and the named groups where the only new things here and you already added tests for them 👍

package.json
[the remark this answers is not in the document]
nfantone: Let's leave like this, then.

nfantone:
Ah, I see. I can move the line in the changelog, that's no problem. But not sure I see what you're referring to as a "feature". More like a change of how things should be done for stuff that was already there. And, for that, all tests have been already accommodated and we [have a couple targeting that change specifically](https://github.com/pillarjs/router/pull/102/files#diff-859f2b16134f796c6f0bb540c54d824024ff05b9722708e13ee78b29476e3471R693). Do you think we need more?

nfantone:
(Welp, I didn't realize I had already committed the `semver` removal change last week - so it went together with the changelog update. Sorry for the confusion - should do the same trick).

nfantone:
@dougwilson Are you happy to merge this? Or is there anything else you'd like to include here?

.github/workflows/ci.yml
```diff
@@ -33,47 +33,47 @@ jobs:
         include:
         - name: Node.js 0.10
           node-version: "0.10"
-          npm-i: mocha@3.5.3 nyc@10.3.2 supertest@2.0.0
+          npm-i: mocha@3.5.3 nyc@10.3.2 semver@6.3.0 supertest@2.0.0
```
nfantone, line 36: @dougwilson This is no longer needed.

nfantone:
@dougwilson Pong?

dougwilson:
Hi, sorry I was out the beginning of this week, and unfortunately trying to now recover from a very bad sickness. I will take a look at what I can as fast as I can.

.github/workflows/ci.yml
dougwilson: You're right. Do you want to remove or have me remove?

HISTORY.md
```diff
@@ -3,35 +3,41 @@
...
     - Named matching groups no longer available by position in `req.params`.
       `/:foo(.*)` only captures as `req.params.foo` and not available as
-      `req.parmas[0]`.
+      `req.params[0]`.
     - Regular expressions can only be used in a matching group.
       `/\\d+` is no longer valid and must be written as `/(\\d+)`.
     - Special `*` path segment behavior removed.
       `/foo/*/bar` will match a literal `*` as the middle segment.
-  * deps: setprototypeof@1.2.0
+    - Parenthesis and bracket literals now require escaping, i.e.: `/user(:id)` or `/user{:id}` no longer match `/  user(42)`, `/user{42}` as before unless written as `/user\\(:id\\)` or `/user\\{:id\\}`.
```
dougwilson, line 31: I think this got messed up in the move. From my recollection from the conversation only the parenthesis stuff changed in this release; the curly brace stuff did change in this PR. Only the parans one should have been moved down here, right?

nfantone:
>  unfortunately trying to now recover from a very bad sickness.

Hope you're feeling better! I'll take a look at your comment above now.

.github/workflows/ci.yml
nfantone: I'll remove it for us.

HISTORY.md
nfantone: @dougwilson I think you're right. Corrected it. Also rebased with latest `2.0`.

nfantone:
@dougwilson Anything else you'd like me to address?

HISTORY.md
```diff
@@ -4,18 +4,23 @@
 This incorporates all changes after 1.3.5 up to 1.3.6.
 
   * Add support for returned, rejected Promises to `router.param`
+  * **BREAKING**: deps: `path-to-regexp@6.2.0` (See [release changelog](https://github.com/pillarjs/path-to-regexp/releases/tag/v6.2.0))
+    - Support for **named capturing groups** in paths using `RegExp`, i.e.: `/\/(?<group>.+)/` would result in `req.params.group` being populated.
+    - Custom **prefix and suffix groups** using `{}`, e.g.: `/:entity{-:action}?` would match `/user` and `/user-delete`.
+    - Unbalanced patterns would now produce an error: `/test(foo` previously would worked, now it expects either `(` to be closed or the character to be escaped for the previous behavior, e.g. `/test\\(foo`.
```
dougwilson, line 10:
```suggestion
    - Unbalanced patterns now produces an error: `/test(foo` previously worked, but now it expects `(` to escaped for the previous behavior, e.g. `/test\\(foo`.
```

I think this wording may be better/clearer. It seems that only escaping it actually reproduces the previous behavior; closing the parans results in different behavior as it now created a capture group.

nfantone: I borrowed the wording straight out of [`path-to-regexp` release notes](https://github.com/pillarjs/path-to-regexp/releases/tag/v6.0.0).

dougwilson: Ok, but the English does not make sense as it is written here. If you have a better suggestion, please feel free to update it.

nfantone: @dougwilson Done. I went with your suggestion. Anything else?

HISTORY.md
```diff
@@ -4,18 +4,23 @@
 This incorporates all changes after 1.3.5 up to 1.3.6.
 
   * Add support for returned, rejected Promises to `router.param`
+  * **BREAKING**: deps: `path-to-regexp@6.2.0` (See [release changelog](https://github.com/pillarjs/path-to-regexp/releases/tag/v6.2.0))
+    - Support for **named capturing groups** in paths using `RegExp`, i.e.: `/\/(?<group>.+)/` would result in `req.params.group` being populated.
+    - Custom **prefix and suffix groups** using `{}`, e.g.: `/:entity{-:action}?` would match `/user` and `/user-delete`.
+    - Unbalanced patterns now produce an error: `/test(foo` previously worked, but now it expects `(` to be escaped for the previous behavior, e.g. `/test\\(foo`.
```
xjamundx, line 10: The release notes referenced above says the previous behavior can be achieved with `/test\(foo` but here it states `/test\\(foo`. Is there a reason for the difference? Apologies if I missed something obvious!

nfantone: Hey @xjamundx 👋🏼 Thanks for chiming in.

You're absolute right, good catch. Yeah, I think `path-to-regexp` release notes are wrong, or misleading at the very least. `/test(foo` no longer works because it starts a capturing group that it's never closed - but if you want the literal `(` to match you'd need to escape it with '\'. This being JS, escaping non-special chars with `\` just echoes the char bar (i.e.: `'\('` -> `'('`). So in order to forward the escaped sequence to the underlying `RegExp`, you'd need to escape the backslash as well.

See it in action:

```js
> const { pathToRegexp } = require('path-to-regexp')

> pathToRegexp('/test\(foo').test('/test(foo')
Uncaught TypeError: Unbalanced pattern at 5

> pathToRegexp('/test\\(foo').test('/test(foo')
true
```

In short, I think `router`'s description makes more sense and should cause less confusion.

HISTORY.md
nfantone: 👀

xjamundx: 🙏🏻

nfantone:
@dougwilson Happy new year! Do you think you could take a look at this PR in the near future? If not, would there be anybody else that could take over from you? Really eager to merge this, help the community and get the release finally going.

HISTORY.md
```diff
@@ -170,8 +176,8 @@ This incorporates all changes after 1.3.3 up to 1.3.5.
   * Fix OPTIONS responses to include the HEAD method properly
   * Remove support for leading colon in `router.param(name, fn)`
   * Use `array-flatten` for flattening arrays
-  * deps: debug@~2.1.1
-  * deps: methods@~1.1.1
+  * deps: `debug@~2.1.1`
+  * deps: `methods@~1.1.1`
```
dougwilson, line 180: Can all these extra unrelated changes be cleaned up or do you want me to review them?

dougwilson:
Hi @nfantone I should be out of the hospital soon. @xjamundx has been helping out.

nfantone:
Hey @dougwilson, sorry for pressuring, I did not know you were hospitalized. Hope you're feeling better.

HISTORY.md
nfantone: Don't think there's much "to review". I don't care about them. I'll revert them.

dougwilson: Ok. Otherwise afaik this pr is good to go. I will merge it asap.

nfantone: Done. `HISTORY.md` should be clean.

euoia:
I have created a pull request in the documentation repo to document these changes in the Express v5 migration guide: https://github.com/expressjs/expressjs.com/pull/1316

nfantone:
@dougwilson Any chance of reviving this? What's the current holdup?

silverwind:
Bump, this seems to be the last blocker for 2.0.0.

hypesystem:
If anyone can do a recap of what exactly needs to be done to land this, I'm very willing to put in the work :smile: But I'll be honest, I find it a bit hard to navigate what's going on right now...

wesleytodd:
Hey @nfantone! [I am the person who landed the first `path-to-regexp` update in this branch](https://github.com/pillarjs/router/pull/42). I am both sad and happy to say that I am working to get the thing lined up to finally wrap up this long road toward landing this in main-line `express`. I wanted to check in with you on this work. If you are willing to work with me I can probably help you land this before our next release. If not, I will likely say we will need to wait for 6.0 (I promise it will not be 10 years). Let me know asap so I can line this up with the plans.

nfantone:
Hey @wesleytodd — happy to help. Honestly, I barely remember what I did here, but seems like it was awaiting a review.

I'll follow your lead. Let me know what I can do to help.

blakeembrey:
Latest should be good. I'd land this, and then maybe another minor that I need to produce to add back `*` support (since it's been standardized in URLPattern now).

Once we have the min node.js version support defined for node 5.x I can also set that up in `path-to-regexp`. Currently I'm producing ES5 code, so the library _probably_ works all the way back to v4 or v5 of node.js, but I'll need to test for that which usually means having to downgrade to a much older test runner (most latest versions only do LTS onward).

wesleytodd:
Hey, so an update on this. We are going a bit of a different direction now. I am probably going to need to subsume this PR into https://github.com/pillarjs/router/pull/117 (where applicaable) because we are entirely dropping the regular expression handling in `path-to-regexp` in favor of only supporting "bring your own regexp" for that subset of features. 

Really sorry this never landed, as the original person to land an update like this many years ago, I feel the disappointment on doing all this work for it not to ever be used by users very keenly. That said, we are moving forward and should hopefully avoid this kind of stagnation in the future.

PR #107, opened 2023-02-28
'continue' is unnecessary as the last statement in while loop

raksbisht:
'continue' is unnecessary as the last statement in while loop inside next() method


wesleytodd (approved):
Looks like this was a result of [this chage](https://github.com/pillarjs/router/commit/e2c97755776a731c25d257dea46195d9ac0673c3). My guess is all `return`s were swapped with `continue`. It is unnecessary I think and since CI passes I think it is good to merge this even though typically I don't like landing changes like this which do not meaningfully do anything.

PR #108, opened 2023-03-28
Add type definition

ilikejames:
A few years ago I had a quick attempt at adding types for this library. [https://github.com/pillarjs/router/pull/76](#76).

This looks over the http methods exposed from the `methods` library and exposes these as lowercase methods on the `Router` return object e.g. `new Router().connect()`

This does mean that `@types/methods` has to be added as a `dependency`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "license": "MIT",
   "repository": "pillarjs/router",
   "dependencies": {
+    "@types/methods": "^1.1.1",
     "array-flatten": "3.0.0",
     "debug": "2.6.9",
     "methods": "~1.1.2",
@@ -18,6 +19,7 @@
     "utils-merge": "1.0.1"
   },
   "devDependencies": {
+    "@types/node": "^18.0.0",
     "after": "0.8.2",
     "eslint": "8.34.0",
     "eslint-plugin-markdown": "3.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -43,6 +47,7 @@
     "test": "mocha --reporter spec --bail --check-leaks test/",
     "test-ci": "nyc --reporter=lcov --reporter=text npm test",
     "test-cov": "nyc --reporter=text npm test",
+    "test-types": "npx -p typescript tsc --project tsconfig.json --noEmit",
```
ilikejames, line 50: Run the typescript checker via `npx`. `typescript` package cannot be listed in  `devDependencies` due to support of node@0.8 which is incompatible.

test/types.ts
```diff
@@ -0,0 +1,81 @@
...
+router.get('/', routerHandler);
+router.post('/', routerHandler);
+router.delete('/', routerHandler);
+router.patch('/', routerHandler);
+router.options('/', routerHandler);
+router.head('/', routerHandler);
+router.bind('/', routerHandler);
+router.connect('/', routerHandler);
+router.trace('/', routerHandler);
+router['m-search']('/', routerHandler);
```
ilikejames, line 34: All http methods exposed from `@types/methods` are exposed on the `router`.

ilikejames:
I took some inspiration from [PR-100](https://github.com/pillarjs/router/pull/100/files) for the build steps.

PR #110, opened 2023-03-28
Add type definition

ilikejames:
A few years ago I had a quick attempt at adding types for this library. [https://github.com/pillarjs/router/pull/76](#76).

This looks over the http methods exposed from the `methods` library and exposes these as lowercase methods on the `Router` return object e.g. `new Router().connect()`

This does mean that `@types/methods` has to be added as a `dependency`,  but it means only the single package will need to be updated as `methods` itself is updated. 

I took some inspiration from [PR-100](https://github.com/pillarjs/router/pull/100) for the build steps. 

(Not sure why previous was not building).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,96 @@
+import Methods from "methods";
+import { OutgoingMessage } from "http";
+
+export default Router;
+
+type Method = typeof Methods[number];
+
+type LowercaseMethods = Lowercase<Method>;
+
+export interface RouterOptions {
+  strict?: boolean;
+  caseSensitive?: boolean;
+  mergeParams?: boolean;
+}
+
+export interface IncomingRequest {
+  url?: string;
+  method?: string;
+  originalUrl?: string;
+  params?: Record<string, unknown>;
+}
+
+export interface RoutedRequest extends IncomingRequest {
+  baseUrl: string;
+  next?: NextFunction;
+  route?: IRoute;
+}
+
+export interface NextFunction {
+  (err?: any): void;
+}
+
+type IRoute = Record<LowercaseMethods, IRouterHandler<IRoute>> & {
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -0,0 +1,122 @@
...
+  'mkactivity' |
+  'checkout' |
+  'merge' |
+  'm-search' |
+  'notify' |
+  'subscribe' |
+  'unsubscribe' |
+  'patch' |
+  'search' |
+  'connect'
```
ilikejames, line 32: Take from [methods lib](https://github.com/jshttp/methods/blob/master/index.js#L40-L67)

index.d.ts
```diff
@@ -0,0 +1,122 @@
...
+  strict?: boolean;
+  caseSensitive?: boolean;
+  mergeParams?: boolean;
+}
+
+export interface IncomingRequest {
+  url?: string;
+  method?: string;
+  originalUrl?: string;
+  params?: Record<string, string>;
```
ilikejames, line 44: Assumption: params are always strings.

package.json
```diff
@@ -18,23 +18,29 @@
...
     "LICENSE",
     "HISTORY.md",
     "README.md",
     "SECURITY.md",
-    "index.js"
+    "index.js",
+    "index.d.ts"
   ],
+  "typesVersions": {
+    ">=4.0": {"*": ["index.d.ts"]}
```
ilikejames, line 42: Limited to downstream `typescript@>4.0.0`

reteps:
Is there any progress here? AFAIK This is the last non-dev dependency to get [Express types moving](https://github.com/expressjs/typescript-wg/issues/6#issuecomment-3267328864).

reteps:
cc @ilikejames

PR #114, opened 2024-03-16
2.0 v1 sync

wesleytodd:
Updating the 2.0 branch. Note, this branch in it's current form does not work with node 20 or `npm install` without `--legacy-peer-deps`. [v1 `router` does get a fair number of downloads]https://www.npmjs.com/package/router?activeTab=versions) and to trial out a bit of a release process change I am going to get that fixed on `master`, but just wanted to call out why this branch is broken at the moment.


wesleytodd:
I am pretty sure this repo is in a bad state. The I think [some of the CICD changes](https://github.com/pillarjs/router/blob/master/.github/workflows/ci.yml) accidentally [dropped some required pin-back functionality](https://github.com/pillarjs/router/commit/4b5db127545bf6264e14eecde4597f34860ad05e) for `supertest` and `nyc`. [I am debugging](https://github.com/pillarjs/router/pull/115) and will try to fix before merging. Like it is there but not appearing to work correctly.

PR #117, opened 2024-08-24
path-to-regexp@^8.0.0

wesleytodd:
Updates to `path-to-regexp@8`.

So from the tests, the three breaking changes are:

1. We drop support for partial capture group `router.route('/user(s?)/:user/:op')` but still have optional non-capture `/user{s}/:user/:op`
2. `:name?` becomes `{:name}`
3. `:name*` becomes `*name`. 
4. The splat change also changes splat from strings to an array of strings
5. Optional splats become `{*name}`
6. `:name+` becomes `*name` and thus equivalent to `*name` so I dropped those tests
7. Strings as regular expressions are fully removed, need to be converted to native regular expressions

I am going to link this into `express` and see locally if this set of breaking changes holds up in that test suite now.

### DIFF
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -20,8 +20,8 @@ const pathRegexp = require('path-to-regexp')
  * @private
  */
 
-const hasOwnProperty = Object.prototype.hasOwnProperty
 const TRAILING_SLASH_REGEXP = /\/+$/
+const MATCHING_GROUP_REGEXP = /\((?:\?<(.*?)>)?(?!\?)/g
 
 /**
  * Expose `Layer`.
@@ -41,10 +41,53 @@ function Layer (path, options, fn) {
   this.name = fn.name || '<anonymous>'
   this.params = undefined
   this.path = undefined
-  this.regexp = pathRegexp((opts.strict ? path : loosen(path)), this.keys, opts)
+  this.slash = path === '/' && opts.end === false
 
-  // set fast path flags
-  this.regexp._slash = path === '/' && opts.end === false
+  function matcher (_path) {
+    if (_path instanceof RegExp) {
+      const keys = []
+      let name = 0
+      let m
+      // eslint-disable-next-line no-cond-assign
+      while (m = MATCHING_GROUP_REGEXP.exec(_path.source)) {
+        keys.push({
+          name: m[1] || name++,
+          offset: m.index
+        })
+      }
+
+      return function regexpMatcher (p) {
+        const match = _path.exec(p)
+        if (!match) {
+          return false
+        }
+
+        const params = {}
+        for (let i = 1; i < match.length; i++) {
+          const key = keys[i - 1]
+          const prop = key.name
+          const val = decodeParam(match[i])
+
+          if (val !== undefined) {
+            params[prop] = val
+          }
+        }
+
+        return {
+          params,
+          path: match[0]
+        }
+      }
+    }
+
+    return pathRegexp.match((opts.strict ? _path : loosen(_path)), {
+      sensitive: opts.sensitive,
+      end: opts.end,
+      trailing: !opts.strict,
+      decode: decodeParam
+    })
+  }
+  this.matchers = Array.isArray(path) ? path.map(matcher) : [matcher(path)]
 }
 
 /**
@@ -126,14 +169,18 @@ Layer.prototype.match = function match (path) {
 
   if (path != null) {
     // fast path non-ending match for / (any path matches)
-    if (this.regexp._slash) {
+    if (this.slash) {
       this.params = {}
       this.path = ''
       return true
     }
 
-    // match the path
-    match = this.regexp.exec(path)
+    let i = 0
+    while (!match && i < this.matchers.length) {
+      // match the path
+      match = this.matchers[i](path)
+      i++
+    }
   }
 
   if (!match) {
@@ -143,22 +190,9 @@ Layer.prototype.match = function match (path) {
   }
 
   // store values
-  this.params = {}
-  this.path = match[0]
-
-  // iterate matches
-  const keys = this.keys
-  const params = this.params
-
-  for (let i = 1; i < match.length; i++) {
-    const key = keys[i - 1]
-    const prop = key.name
-    const val = decodeParam(match[i])
-
-    if (val !== undefined || !(hasOwnProperty.call(params, prop))) {
-      params[prop] = val
-    }
-  }
+  this.params = match.params
+  this.path = match.path
+  this.keys = Object.keys(match.params)
 
   return true
 }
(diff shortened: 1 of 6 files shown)


lib/layer.js
```diff
@@ -41,7 +40,7 @@ function Layer (path, options, fn) {
   this.name = fn.name || '<anonymous>'
   this.params = undefined
   this.path = undefined
-  this.regexp = pathRegexp((opts.strict ? path : loosen(path)), this.keys, opts)
+  this.regexp = pathRegexp.match((opts.strict ? path : loosen(path)), opts)
```
blakeembrey, line 43: Can we get rid of the `_slash` behavior? If it's a real perf win I'll move it into `path-to-regexp`?

blakeembrey, line 43: May also want to pass through `decode: decodeParam` if you want the 400 error behavior, the default `decode = decodeURIComponent` so there's no wrapper.

wesleytodd: Yeah that would be great. I am not sure it is a perf win anymore now that we have removed the regexp parts. For now, all the tests are passing locally (after the updates I just pushed) other than the array ones. Once you get that added, we can probably publish since the `_slash` optimization can be changed in a minor.

lib/layer.js
```diff
@@ -41,10 +40,27 @@ function Layer (path, options, fn) {
...
+  var matchers = []
+
+  function addMatcher (_path) {
+    var matcher = pathRegexp.match((opts.strict ? _path : loosen(_path)), {
+      sensitive: opts.sensitive,
+      end: opts.end,
+      trailing: !opts.strict,
+      decode: decodeParam
+    })
+    matchers.push(matcher)
```
blakeembrey, line 54: Stylistic, but not a huge fan of code with side effects. I think this would be better written as: 

```js
function matcher(path) { return pathToRegexp.match(...); }

const matchers = Array.isArray(path) ? path.map(matcher) : [matcher(path)];
```

lib/layer.js
wesleytodd: So while I was working through the other test fixes after this I figured out what it was that fixed.

```
const fn = match('/', {
  end: false
});
console.log(fn('/foo'));
```

This case is what `.use` does, and it means that the `.use(mw)` call does not match the route. To me this behavior just seems broken and like we could replace it with much more simple ways, but for now I am trying to do the minimal change set to get the update done. I brought it back after trying twice to delete it lol, so was not minimal in the amount of work, but it reduces the changes in this PR a bit.

Are we alright with this compromise for now until we can get the release out and then do more fixes in minors?

lib/layer.js
wesleytodd: I also did not like this code lol, but it was my first attempt and it worked so I was tired so took a nap instead of changing it. But I agree and will update it.

lib/layer.js
```diff
@@ -41,10 +41,53 @@ function Layer (path, options, fn) {
...
-  // set fast path flags
-  this.regexp._slash = path === '/' && opts.end === false
+    return pathRegexp.match((opts.strict ? _path : loosen(_path)), {
+      sensitive: opts.sensitive,
+      end: opts.end,
+      trailing: !opts.strict,
+      decode: decodeParam
+    })
+  }
+  this.matchers = Array.isArray(path) ? path.map(matcher) : [matcher(path)]
```
blakeembrey, line 90: I would personally recommend forwarding the array as-is through to `path-to-regexp` now, and only support non-array regexes. Mostly because I intent to work as much as possible on optimizing the matching code outside of router when possible. I.e. the path signature is `string | string[] | RegExp`.

lib/layer.js
```diff
@@ -41,10 +41,53 @@ function Layer (path, options, fn) {
...
+          return false
+        }
+
+        var params = {}
+        for (var i = 1; i < match.length; i++) {
+          var key = keys[i - 1]
+          var prop = key.name
+          var val = decodeParam(match[i])
+
+          if (val !== undefined || !(hasOwnProperty.call(params, prop))) {
```
blakeembrey, line 71: I think you can get rid of the `hasOwnProperty`? Last write wins has been how it worked forever already.

blakeembrey (approved):

lib/layer.js
wesleytodd: Oh I forgot we talked about that in slack but it was after I had added this here. I will remove that.

lib/layer.js
wesleytodd: ah yeah, that was just copy pasta.

wesleytodd:
I am about to rebase this onto the new CI and lint changes. I should probably just life out the changes, but I like a challenge with git and what is the point of doing this if we are not having fun while we do it?

wesleytodd:
ok I think I did this correctly. There was a duplicate test and a few things which made it confusing, but I think once we get `path-to-regexp@8` released this is nearly ready.

test/router.js
```diff
@@ -400,7 +372,7 @@ describe('Router', function () {
               .end(cb)
           },
           function (cb) {
-            request(server)[method]('/zoo')
+            request(server)[method]('/bar')
```
wesleytodd, line 375: I was trying to figure out how this one worked before, but I am thinking this might be a weird git diff thing from a rebase. The `/bar` route should not have "hit" it would have needed the line 2 below to flip if we kept this as `zoo`. Anyway, not a really meaningful change so I am not going to try and undo what ever I missed in the merge and just push forward.

test/router.js
```diff
@@ -1049,8 +1070,8 @@ describe('Router', function () {
 
       request(server)
         .get('/test/api/1234')
-        .expect(shouldNotHitHandle(1))
-        .expect(shouldNotHitHandle(2))
+        .expect(shouldHitHandle(1))
+        .expect(shouldHitHandle(2))
```
wesleytodd, line 1074: Looks like this change was because something was broken it just is not covered in these tests but is covered in the express tests. Will try and port over the correct test here.

test/route.js
```diff
@@ -716,27 +716,6 @@ describe('Router', function () {
             .expect(200, { foo: 'fizz', bar: 'buzz' }, done)
         })
 
-        it('should work following a partial capture group', function (done) {
```
wesleytodd, line 719: While looking at the `express` tests I think we might be able to bring this one back in a different form. The interesting part is that the optional group is no longer captured because we dropped support for unnamed params. @blakeembrey just want to confirm this is our expected behavior.

This is the test as written in `express`:

```
    it('should work following a partial capture group', function(done){
      var app = express();
      var cb = after(2, done);

      app.get('/user{s}/:user/:op', function(req, res){
        res.end(req.params.op + 'ing ' + req.params.user + (req.url.startsWith('/users') ? ' (old)' : ''));
      });

      request(app)
      .get('/user/tj/edit')
      .expect('editing tj', cb);

      request(app)
      .get('/users/tj/edit')
      .expect('editing tj (old)', cb);
    })
```

blakeembrey: This is correct, there's no regex or unnamed group features right now. I think we could bring this back given time but right now it isn't possible.

PR #119, opened 2024-09-01
Drop support for node <18

wesleytodd:


IamLizu (approved):

wesleytodd:
I saw this test fail locally in some cases. Will look into it now that I also see it in CI. So many of these tests are failing locally but start working when I seralize the requests. I am nearly positive this is a superagent bug, I will pull over all the fixes I was doing in #117 using `serial` I guess.

wesleytodd:
Ah `QUERY` my old friend. I guess this never got fixed in node 21. I am going to exclude that test for node 21 for now, and since that is not an LTS we can just remove it when we drop support for node 21 I think.

PR #123, opened 2024-09-25
refactor: initialize `this.path` with `path`

### LINKED ISSUE #122: Why is `this.path` set to undefined instead of `path`?
Can someone please help me understand why `this.path` is set to `undefined`?
https://github.com/pillarjs/router/blob/2e7fb67ad1b0c1cd2d9eb35c2244439c5c57891a/lib/layer.js#L43

I had checked the router _Layer_, and the path is was `undefined` in almost every route, that shouldn't be the case, right?

While I don't know the reason of it being set to `undefined` initially, I can suggest it should be set to `path`. I have noticed that the `path` on _Layer_ is no longer undefined if we set,
```js
this.path = path;
```

IamLizu:
- Fixes #122

### DIFF
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -40,7 +40,7 @@ function Layer (path, options, fn) {
   this.keys = []
   this.name = fn.name || '<anonymous>'
   this.params = undefined
-  this.path = undefined
+  this.path = path
   this.slash = path === '/' && opts.end === false
 
   function matcher (_path) {


lib/layer.js
```diff
@@ -40,7 +40,7 @@ function Layer (path, options, fn) {
   this.keys = []
   this.name = fn.name || '<anonymous>'
   this.params = undefined
-  this.path = undefined
+  this.path = path
```
carpasse, line 43: I think this property is expected to contain the matched `path` and not the possible array of paths of the

wesleytodd: Agreed, IIRC the idea here is that it is `undefined` until matched, then is also unset. You can see this in `.match`. I am pretty sure that means this would break the intended behavior.

wesleytodd (changes requested):

IamLizu:
Thanks for clarifying the intent here. I’ll go ahead and close this for now.

PR #125, opened 2024-10-21
Remove Object.setPrototypeOf polyfill

Phillip9587:
This PR removes the `Object.setPrototypeOf` polyfill, which is no longer needed since this package now requires Node.js v18 as the minimum supported version. All existing tests should pass, as this change only impacts older Node versions no longer supported.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+==================
+
+* Remove `Object.setPrototypeOf` polyfill
+
 2.0.0 / 2024-09-09
 ==================
 
--- a/index.js
+++ b/index.js
@@ -19,7 +19,6 @@ const methods = require('methods')
 const mixin = require('utils-merge')
 const parseUrl = require('parseurl')
 const Route = require('./lib/route')
-const setPrototypeOf = require('setprototypeof')
 
 /**
  * Module variables.
@@ -60,7 +59,7 @@ function Router (options) {
   }
 
   // inherit from the correct prototype
-  setPrototypeOf(router, this)
+  Object.setPrototypeOf(router, this)
 
   router.caseSensitive = opts.caseSensitive
   router.mergeParams = opts.mergeParams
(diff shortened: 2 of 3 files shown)


jonkoops (approved):

UlisesGascon (approved):
LGTM!

wesleytodd:
Hey, I know I have been slow to review things but I wonder if we could keep things open for a bit longer than a day? This was added originally to support *browsers* and we probably should have had a browser compatibility support listed somewhere and ideally tested, but it was a long time ago and I don't remember the discussions from back then. But to me this should have been considered a breaking change unless we were also going to consider 2.x having dropped browser supports. If we want that (which I think is a good idea) then we should also document this change as if it was part of what we dropped in 2.x.

UlisesGascon:
> but I wonder if we could keep things open for a bit longer than a day?

Yep, l will work on small policy for this and other stuff related to PRs.

>  This was added originally to support browsers and we probably should have had a browser compatibility support listed somewhere and ideally tested, but it was a long time ago and I don't remember the discussions from back then (...)

In this case I think that we are safe as `setPrototypeOf` is super supported since a long time ago https://caniuse.com/mdn-javascript_builtins_object_setprototypeof.

I think that following with the proposal of this policy (https://github.com/expressjs/discussions/pull/289) we should include tests for the things that we actually maintain. So, if the library was "unintended" working fine for the browsers probably we want to include tests for this otherwise "unintended" features might break at any time. Also this applies for any runtime that we don't specifically support, in this scenario I considered the browsers in this category.

wesleytodd:
> In this case I think that we are safe as setPrototypeOf is super supported since a long time ago

Yeah, I totally agree and removing it (and all the other unneeded polyfill type things) is a great thing. I was just being a stickler for semver 🤓.

---

I think we can continue the browser support discussion in #128 and then revisit this once we make a decision there?

PR #126, opened 2024-10-21
Use `Array.flat` instead of `array-flatten` package

Phillip9587:
Since the mimimum supported Node version is v18 the native `Array.flat` method should be used instead of the `array-flatten` package.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+=================
+
+* Use `Array.flat` instead of `array-flatten` package
+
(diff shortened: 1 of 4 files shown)


jonkoops (approved):
Not a project maintainer, but looking to help clean up the repo and this looks good to me.

UlisesGascon (approved):
LGTM!

UlisesGascon:
We probably want to reflect this change in the `HISTORY.md` file.

PR #127, opened 2024-10-21
refactor: replace `methods` dependency with standard library

jonkoops:
Replaces the [`methods`](https://github.com/jshttp/methods/tree/master) dependency with calls to the standard library of Node.js. Since `methods` is mostly a polyfill for `require('node:http').METHODS` in older versions of Node.js that are no longer supported this can be removed.


blakeembrey (approved):

wesleytodd (approved):
And another one bites the dust!

PR #137, opened 2024-10-27
refactor!: replace `is-promise` dependency with `instanceof` check

### LINKED ISSUE #136: Consider dropping `is-promise` as a dependency
Currently this module depends on [`is-promise`](https://github.com/then/is-promise), which allows 'promise like' objects with a `then()` method to be interpreted as a promise, for example:

```js
isPromise({then:function () {...}}); // true
```

This has been historically done to handle polyfilling Promises in environments where they were not yet fully supported (e.g. the `Promise` global is not available).  Since this is no longer needed as promises are widely supported it might be time to drop this dependency and instead use a simple `instanceof Promise` check, which would only allow standardized promises.

jonkoops:
Replaces the `is-promise` dependency with an `instanceof Promise` check.

BREAKING: drops support for non-native promises

Closes #136

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,7 @@ unreleased
 * Remove `Object.setPrototypeOf` polyfill
 * Use `Array.flat` instead of `array-flatten` package
 * Replace `methods` dependency with standard library
+* Replace `is-promise` dependency with `instanceof` check (drops support for non-native promises)
 
 2.0.0 / 2024-09-09
 ==================
--- a/index.js
+++ b/index.js
@@ -12,7 +12,6 @@
  * @private
  */
 
-const isPromise = require('is-promise')
 const Layer = require('./lib/layer')
 const methods = require('methods')
 const mixin = require('utils-merge')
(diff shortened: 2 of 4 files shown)


jonkoops:
This PR should likely not be merged until the next major.

jonkoops:
Just ping me whenever a new major is in the works and I will take a moment to rebase the work here.

wesleytodd:
Really sorry to have left this one hanging for so long. You are right this is a breaking chance since it would drop support for some of the [legacy promise forms that package supports](https://github.com/then/is-promise?tab=readme-ov-file#api). And looking back at the timing we had already cut the 2.x release line by the time this was opened anyway.

So, what can we do? I think the best option we could do is to check if it is anything *other than a native `Promise`* and print a deprecation warning in the 2.x branch. That would set us up to remove this package and fully drop support for anything but native promises in the 3.x line which will ship with `express@6`. Does that sound like a good next step to you?

jonkoops:
Sure, I can go ahead and make a PR that logs a deprecation for this. Do you have any examples of how deprecations should be logged that I could use as inspiration?

bjohansebas:
I like the idea of showing that warning

jonkoops:
Created a PR to add the message under #154

PR #140, opened 2024-11-05
cleanup: remove devDependency `safe-buffer`

Phillip9587:
This PR removes the `safe-buffer` package from `devDependencies` as it's no longer needed. Since we no longer support Node.js versions older than 18.0.0, the native Buffer class now fully handles security concerns. Additionally, there are no instances of `new Buffer(...)` usage in the repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "mocha": "10.2.0",
     "nyc": "15.1.0",
     "run-series": "^1.1.9",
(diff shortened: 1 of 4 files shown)


jonkoops (approved):

bjohansebas (approved):

UlisesGascon (approved):

bjohansebas:
I thought this had already been released

PR #143, opened 2024-11-05
ci: updated github actions ci workflow

### LINKED ISSUE #142: Update Github Actions Workflow
I noticed that the current CI workflow in this repository could benefit from some updates. Specifically:

1. **Node.js Installation in the Matrix Test Run:**
   - Currently, the workflow uses `nvm` to install Node.js versions. 
   - It might be worth considering switching to the official [`actions/setup-node`](https://github.com/actions/setup-node) action. This action can leverage cached Node.js versions from the runner, which could improve efficiency and speed up the CI pipeline.

2. **Deprecation of Artifact Actions v3:**
   - The `actions/upload-artifact@v3` and `actions/download-artifact@v3` actions are being deprecated as of **November 30, 2024** ([GitHub Deprecation Notice](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)).
   - These actions should be updated to their latest versions to ensure continued functionality in the CI workflow.

3. **The Coverage setup could also be optimized**:
    - Currently, the workflow uses the `coverallsapp/github-action@master` which points to v1 of this action. This v1 action uses `node16` as runtime which is deprecated.

4. **Minimum token permissions for the GITHUB_TOKEN**:
    - GitHub Actions workflows have a GITHUB_TOKEN with write access to multiple scopes. 
    - It is a best practice to reduce the scopes to the minimum needed for each workflow / step.
    - GitHub recommends defining minimum GITHUB_TOKEN permissions. https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
    - The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Phillip9587:
closes #142 

The same has been done for `body-parser`: https://github.com/expressjs/body-parser/pull/546

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,126 +1,74 @@
 name: ci
-
 on:
   push:
     branches:
       - master
-      - '2.0'
     paths-ignore:
       - '*.md'
   pull_request:
     paths-ignore:
       - '*.md'
 
+permissions:
+  contents: read
+
 jobs:
   test:
+    name: Test - Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        name:
-        - Node.js 18.x
-        - Node.js 19.x
-        - Node.js 20.x
-        - Node.js 21.x
-        - Node.js 22.x
-
-        include:
-        - name: Node.js 18.x
-          node-version: "18"
-
-        - name: Node.js 19.x
-          node-version: "19"
-
-        - name: Node.js 20.x
-          node-version: "20"
-
-        - name: Node.js 21.x
-          node-version: "21"
-
-        - name: Node.js 22.x
-          node-version: "22"
+        node-version: [18, 19, 20, 21, 22, 23]
 
     steps:
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
-    - name: Install Node.js ${{ matrix.node-version }}
-      shell: bash -eo pipefail -l {0}
-      run: |
-        nvm install --default ${{ matrix.node-version }}
-        dirname "$(nvm which ${{ matrix.node-version }})" >> "$GITHUB_PATH"
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          check-latest: true
 
-    - name: Configure npm
-      run: |
-        if [[ "$(npm config get package-lock)" == "true" ]]; then
-          npm config set package-lock false
-        else
-          npm config set shrinkwrap false
-        fi
+      - name: Install Node.js dependencies
+        run: npm install
 
-    - name: Install Node.js dependencies
-      run: npm install
+      - name: Run tests
+        run: npm run test-ci
 
-    - name: List environment
-      id: list_env
-      shell: bash
-      run: |
-        echo "node@$(node -v)"
-        echo "npm@$(npm -v)"
-        npm -s ls ||:
-        (npm -s ls --depth=0 ||:) | awk -F'[ @]' 'NR>1 && $2 { print $2 "=" $3 }' >> "$GITHUB_OUTPUT"
+      - name: Lint code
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,126 +1,74 @@
...
-        - name: Node.js 21.x
-          node-version: "21"
-
-        - name: Node.js 22.x
-          node-version: "22"
+        node-version: [18, 19, 20, 21, 22, 23]
 
     steps:
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
```
bjohansebas, line 25:
```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```

It would be better to use the hash of the last commit of the version we want to use in each GitHub action, it's just to avoid opening another PR later.

Example: https://github.com/expressjs/compression/pull/205

Phillip9587: I don't recommend pinning GitHub actions with commit hashes, as it introduces a maintenance burden. Without an automatic dependency update system in place, manually updating these pinned hashes becomes cumbersome. In my opinion, using version tags like v4 ensures stability while allowing for easier maintenance and updates.

Phillip9587: As soon as a dependency update tool is implemented, I fully support pinning to commit hashes to ensure more controlled and reliable updates.

bjohansebas: I would like to know your opinions @UlisesGascon @wesleytodd

jonkoops: I would not bother pinning any actions from GitHub itself. The security risk there is basically non-existent. If those were ever compromised then that would be a world news level exploit. I am pretty sure the only reason this is recommend is that 3rd party actions are much more susceptible, and GitHub wants to cover their bases, hence the recommendation in the docs.

wesleytodd: We need to do an RFC to align on this I think. We have typically considered the core platform ones "safe" in this reagard and the third party ones "unsafe". That said, I think we should have a more serious conversation on the pros/cons for the future. For example, I am strongly opposed to locking unless we have an auto update mechanism in place, ideally with some simple auto-merge logic (passed builds and one approval?).

For today, I think it is good to keep the ranges because that is also what we are doing in most repos.

wesleytodd (changes requested):
I am all 👍 on this, but put request changes so we can decide if we want to address the 1.x thing in this PR.

.github/workflows/ci.yml
```diff
@@ -1,126 +1,74 @@
 name: ci
-
 on:
   push:
     branches:
       - master
-      - '2.0'
```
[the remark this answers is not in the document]
Phillip9587: @wesleytodd I think we should leave the v1 pipeline as it is in a separate 1.x branch. The 1.x branch needs to be tested on node.js version beginning with 0.8 while the master / 2.x branch does not support versions below v18. If we leave the 2 pipelines separate we just need to backport new node.js versions for 1.x.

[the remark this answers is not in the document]
wesleytodd: Ah good point. I think that is actually what we did other places. I was not thinking about that when I made this comment.

wesleytodd (approved):

bjohansebas (approved):

UlisesGascon (approved):

UlisesGascon:
Great point @wesleytodd!

I think that @inigomarquinez is working on several RFC(s) related to the OpenSSF Scorecard adoption that will cover how the locking + upgrade mechanism should work.

.github/workflows/ci.yml
wesleytodd: Lets call this resolved and move the conversation into an RFC so we can merge this improvement.

PR #151, opened 2025-02-17
fix: restore 'debug' logs

dpopp07:
Revert "Remove debug dependency", with the following changes:

- Update "var" to "const" for debug imports
- Preserve history file so intermediate versions are documented correctly
- Use latest version of debug

This reverts commit 0b5a0a6b578565cc36712f8d5004566bf5e1472f.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+==================
+
+* Restore `debug` dependency
+
 2.1.0 / 2025-02-10
 ==================
 
--- a/index.js
+++ b/index.js
@@ -17,6 +17,7 @@ const Layer = require('./lib/layer')
 const { METHODS } = require('node:http')
 const parseUrl = require('parseurl')
 const Route = require('./lib/route')
+const debug = require('debug')('router')
 
 /**
  * Module variables.
@@ -148,6 +149,8 @@ Router.prototype.handle = function handle (req, res, callback) {
   if (!callback) {
     throw new TypeError('argument callback is required')
   }
+
+  debug('dispatching %s %s', req.method, req.url)
 
   let idx = 0
   let methods
@@ -315,6 +318,7 @@ Router.prototype.handle = function handle (req, res, callback) {
 
       // Trim off the part of the url that matches the route
       // middleware (.use stuff) needs to have the path stripped
+      debug('trim prefix (%s) from url %s', layerPath, req.url)
       removed = layerPath
       req.url = protohost + req.url.slice(protohost.length + removed.length)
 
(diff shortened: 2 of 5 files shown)


bjohansebas (approved):
LGTM, please add it to the history

dpopp07:
Note: as-is, this does NOT restore the ability to use `DEBUG=express:*` for viewing router logs. I don't think we should lose the scope of this package, since it's decoupled from Express.

If it was really important to restore that behavior, we could do something hacky like:
```js
const debugCurrent = require('debug')('router')
const debugLegacy = require('debug')('express:router')

function debug(...args) {
  debugCurrent(...args)
  debugLegacy(...args)
}
```

I tested this and it seems to work fine. That said, I'd prefer not to add something like this.

I would rather open a follow-on PR in the v5 migration guide to explicitly state that router logs must now be viewed through the `router` scope (`DEBUG=express:*` -> `DEBUG=express:*,router`).

Let me know what y'all think. I'm open to other options as well.

dpopp07:
> please add it to the history

Done! I thought that might need to wait until the release, forgot about the "unreleased" pattern. Thanks for pointing it out.

dpopp07:
Re: [my prior comment](https://github.com/pillarjs/router/pull/151#issuecomment-2664249415)

I realized that _if_ it's important to restore `express:router` debug logs and if we want to move in the direction of diagnostic channels, we could go ahead and implement diagnostic channels in this library that simply duplicate the current debug logs.

In Express, we could subscribe to this channel, re-print any messages with `debug` under the `express:router` namespace, and it should be just like old times. There may be hidden complexity and I haven't tested it but I think it should work.

@jonchurch this might have been what you suggested the other day, unless you were just proposing that diagnostic channels would be a better approach for the future. Either way, I'm curious what folks think about this.

bjohansebas:
I think diagnostic channels would be for a different PR, and technically this is a separate package, so it doesn’t make sense for debug to be used for `express:router`. This is not a blocking comment, though, in case another member suggests that it should observe `express:router`.

bjohansebas:
At this point, the best option is to update the migration guide about the fact that express:router in debug no longer exists, since it was a breaking change and no longer exists in express@5.0.0.

The diagnostic channel would be great to try in the future.

dpopp07:
I agree. Unless there are any objections before I get to it, I'll go ahead and open that PR against the migration guide

dpopp07:
I opened the follow up PR to the migration guide: https://github.com/expressjs/expressjs.com/pull/1819

This PR will need to be merged and released before that documentation is accurate.

wesleytodd (approved):
Awesome! I don't think we should bring back `express:*` to include `router`, but it might mean we need to add some migration docs on this right? This is technically a breaking change we did already by removing it, so adding it here as the `router` scope is pretty reasonable and forward moving imo.

dpopp07:
Sounds good!

> it might mean we need to add some migration docs on this right?

Yes - I opened a draft PR updating the migration docs. See [my previous comment](https://github.com/pillarjs/router/pull/151#issuecomment-2682499183) 🙂

Issue #162, opened 2025-05-18
Test failing on Node v20.x


bjohansebas:
If someone could help us resolve this, that would be great. (ref: https://github.com/pillarjs/router/actions/runs/15097652965/job/42434464916?pr=161)

nathanhleung:
Investigating — it seems like something changed in the [latest Node 20 release](https://github.com/nodejs/node/releases/tag/v20.19.2) which came out last week

```
nvm install 20.19.1
npm run test

[...]

678 passing (1s)
```

```
nvm install 20.19.2
npm run test

[...]

214 passing (374ms)
  1 failing

  1) Router
       .route(path)
         .query(...fn)
           should respond to a QUERY request:
     Error: expected 200 "OK", got 404 "Not Found"
      at Context.<anonymous> (test/route.js:274:14)
```

nathanhleung:
This is interesting — I made a file `test.js`:

```js
const http = require("http");

const server = http.createServer((req, res) => {
  res.writeHead(200, { "Content-Type": "text/plain" });
  res.end("Hello World\n");
});

server.listen(3000, async () => {
  console.log("Server running at http://localhost:3000/");

  const res = await fetch("http://localhost:3000/", {
    method: "QUERY",
  });
  console.log(res.status);
});
```

I get different output on Node v20.19.1 vs v20.19.2:

```bash
nvm use 20.19.1
node test

Server running at http://localhost:3000/
400
```

```bash
nvm use 20.19.2
node test

Server running at http://localhost:3000/
200
```

nathanhleung:
More testing in `test.js`:

```
const { METHODS } = require("http");

console.log(METHODS);
```

Note that `QUERY` is added in v20.19.2:

```
nvm use 20.19.1
node test

[
  'ACL',         'BIND',       'CHECKOUT',
  'CONNECT',     'COPY',       'DELETE',
  'GET',         'HEAD',       'LINK',
  'LOCK',        'M-SEARCH',   'MERGE',
  'MKACTIVITY',  'MKCALENDAR', 'MKCOL',
  'MOVE',        'NOTIFY',     'OPTIONS',
  'PATCH',       'POST',       'PROPFIND',
  'PROPPATCH',   'PURGE',      'PUT',
  'REBIND',      'REPORT',     'SEARCH',
  'SOURCE',      'SUBSCRIBE',  'TRACE',
  'UNBIND',      'UNLINK',     'UNLOCK',
  'UNSUBSCRIBE'
]
[...]

PR #160, opened 2025-05-16
feat: create `router.error` method to allow explicit definition of error-handling middleware

nathanhleung:
Related issues/PRs: https://github.com/expressjs/express/issues/2896, https://github.com/pillarjs/router/pull/59

# Summary

This PR creates a `router.error` method which enables the explicit definition of error-handling middleware (first proposed by @feross ~10 years ago in https://github.com/expressjs/express/issues/2896) as opposed to implicit definition by passing an arity-4 function to `router.use`.

The new `router.error` method supports error handling middleware defined with either 3 or 4 arguments. The first argument to the middleware will always be the error `err` (thus avoiding the arity gotcha with `router.use`).

```js
// This works, even though the arity of the handler is 3!
router.error(function (err, req, res) {
  res.statusCode = 500
  res.setHeader('Content-Type', 'text/plain')
  res.end('ouch on ' + req.method + ' ' + req.url + ': ' + err.message)
})

// The usual signature also works
router.error(function (err, req, res, next) {
  res.statusCode = 500
  res.setHeader('Content-Type', 'text/plain')
  res.end('ouch on ' + req.method + ' ' + req.url + ': ' + err.message)
})
```

The PR is backwards-compatible and makes no changes to `router.use`. All tests pass locally for me, at least (macOS Sequoia, Node v22).

Let me know your thoughts/feedback on the current approach and can revise.

# Changes

The diff is ~100 source lines and ~600 test lines. The new `router.error` method mostly falls back on the existing logic in `router.use` when possible. Effectively, it wraps its argument in another wrapper function with arity 4 to trigger the existing error handling behavior in `use`, and then the wrapped function is passed to `use`.

The PR differs from https://github.com/pillarjs/router/pull/59 in that the changes are isolated to the router itself, rather than being on the individual route objects.

# Demo

Live demo on CodeSandbox (using my forked version of `router`): https://codesandbox.io/p/devbox/y63xfg?file=%2Findex.js

```js
function logErrorToConsoleMiddleware(err, req, res, next) {
  console.error(err)
  next(err)
}

function logErrorToFileMiddleware(err, req, res, next) {
  fs.appendFile('errors.txt', err.toString(), function() {
    next(err)
  })
}

// Note that the arity of the error handler can be 3 when passed to
// `router.error` (it must be 4 when passed to to `router.use`).
function sendErrorResponseMiddleware(err, req, res) {
  res.statusCode = 500
  res.setHeader('Content-Type', 'text/plain')
[...]

PR #169, opened 2025-06-29
test: skip query test on node <22

nathanhleung:
Follow-up to https://github.com/pillarjs/router/issues/162#issuecomment-3016193694

> This is due to the latest Node.js security patch, it's probably enough to just skip that test
>
> https://github.com/expressjs/express/pull/6512

This mirrors the Express PR (https://github.com/expressjs/express/pull/6512) and skips testing the `QUERY` method on Node <22 in this repository too.

### DIFF
--- a/test/route.js
+++ b/test/route.js
@@ -11,6 +11,7 @@ const shouldHaveBody = utils.shouldHaveBody
 const shouldHitHandle = utils.shouldHitHandle
 const shouldNotHaveBody = utils.shouldNotHaveBody
 const shouldNotHitHandle = utils.shouldNotHitHandle
+const shouldSkipQueryHttpMethod = utils.shouldSkipQueryHttpMethod
 const methods = utils.methods
 
 describe('Router', function () {
@@ -255,9 +256,7 @@ describe('Router', function () {
         return
       }
 
-      // Skipping HTTP QUERY tests below Node 22, QUERY wasn't fully supported by Node until 22
-      const majorVersion = Number(process.versions.node.split('.')[0]);
-      if (method === 'query' && majorVersion < 22) {
+      if (shouldSkipQueryHttpMethod()) {
         return
       }
 
--- a/test/route.js
+++ b/test/route.js
@@ -254,7 +254,10 @@ describe('Router', function () {
         // CONNECT is tricky and supertest doesn't support it
         return
       }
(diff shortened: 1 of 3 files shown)


bjohansebas:
Could you do something like this in Express, please, so we don't have to repeat the same logic multiple times?

https://github.com/jonchurch/express/blob/bd302bf4606259a7703a06a2c5d269eed1679eb1/test/support/utils.js#L78

nathanhleung:
> Could you do something like this in Express, please, so we don't have to repeat the same logic multiple times?
> 
> [jonchurch/express@`bd302bf`/test/support/utils.js#L78](https://github.com/jonchurch/express/blob/bd302bf4606259a7703a06a2c5d269eed1679eb1/test/support/utils.js#L78)

@bjohansebas done, put all the logic into a self-contained function. Let me know if your preference is to match the implementation in Express where the version string is provided by the caller and I can update

bjohansebas:
Now we just have to wait for a maintainer (TC) to approve the CI run, and then I’ll be able to give the green light.

UlisesGascon (approved):

nathanhleung:
cc @UlisesGascon would be great to get this merged — anything I can do to help?

bjohansebas:
Something curious is that the error no longer appears, new PRs were opened and didn’t trigger any error in the CI

nathanhleung:
@bjohansebas oh, that's interesting

i just rebased my original PR https://github.com/pillarjs/router/pull/160 on the latest `master` here, could you approve my workflows and we can see if CI passes?

wesleytodd:
I also did not find any evidence of Node.js removing `query` either, yet the tests pass for node 20. Honestly it was long enough ago now that I am questioning if we ever saw this present in the Router tests. It is possible we only saw the test failures in `express`. 

https://github.com/pillarjs/router/actions/runs/16882212279/job/47820666814?pr=177

Screenshot 2025-08-11 at 8 57 49 AM

Side note: please do not "ping" or "cc" me. I watch all the repo's I can and this makes it take longer to get through issues and buries important issues behind hundreds of noisy notifications. While it is great for folks to keep things moving forward this noise just makes a the problem worse. Thanks!

nathanhleung:
@wesleytodd great, thanks for the follow up. Looks like tests are passing all around without this PR. In that case, are we good to close this?

PR #174, opened 2025-07-28
feat: add getRoutes method to list registered routes

### LINKED ISSUE #122: Why is `this.path` set to undefined instead of `path`?

bjohansebas:
There is a real need to be able to list routes correctly, as clearly shown in issue (https://github.com/AlbertoFdzM/express-list-endpoints/issues/99, https://github.com/expressjs/express/discussions/5961, https://github.com/expressjs/express/discussions/5858, https://github.com/expressjs/express/issues/6481, https://github.com/wesleytodd/express-openapi/blob/main/lib/generate-doc.js#L16-L74, https://github.com/pillarjs/router/issues/122).
Additionally, I’ve been chatting with a Vercel employee who wants to make it possible to deploy Express servers on Vercel without any special configuration, the only blocker for them so far is that they can’t retrieve a list of the application's routes.

Previously, before the upgrade to `path-to-regexp@8` (#117), it was possible to access the regex used for routing. However, this was replaced by the `matcher` property, which returns a function that extracts the path and parameters. This approach adds extra processing overhead, as it transforms a regex into a function—something unnecessary for such a simple goal. That said, it could still serve as an option for those attempting to list routes.

Currently, route layers have a `path` property, but it's initialized as `undefined` because the actual path is only resolved at runtime. This is due to the fact that `req.url.pathname` isn't defined until the request is received. Once the request arrives and the pathname is parsed (as shown in https://github.com/pillarjs/router/blob/3d574ed80df980eec937bd0c842e8cbf3fbf45a6/index.js#L237), the `path` property in the layer is then set with the correct value.

#### Proposed Solution

The solution would be to add a new property to the layer that gets initialized at the time the layer is created, using the path it was registered with. This would allow us to reliably retrieve the original path and make it possible for external tools or modules to list routes correctly.

Additionally, I'm proposing a new function that exposes the current stack's route definitions. This would reduce the need for third-party dependencies,or at least make it easier for those dependencies to support the new version of Express without relying on brittle workarounds.

Examples: 
```javascript
const router = new Router({ strict: true, caseSensitive: true })
const admin = new Router({ strict: true, caseSensitive: false })

admin.use((req, res, next) => {
  // some middleware for admin routes
  next()
})

admin.get('/', (req, res, next) => {
  res.end('Hello')
})

router.use("/admin", admin)

router.all('/:id', function (req, res) {
  res.end('Hello')
})

console.log(router.getRoutes())
// [
//   {
//     name: 'router',
//     path: '/admin',
//     methods: undefined,
//     keys: undefined,
//     router: [
//       {
//         name: 'handle',
//         path: '/',
//         methods: ['GET'],
//         keys: undefined,
//         router: undefined,
//         options: { strict: true, caseSensitive: false, end: true },
//       }
//     ],
//     options: { strict: true, caseSensitive: true, end: false }
//   },
//   {
//     name: 'handle',
//     path: '/:id',
//     methods: ['_ALL'],
//     keys: [{ name: 'id', type: "param" }],
//     router: undefined,
//     options: { strict: true, caseSensitive: true, end: true }
//   }
// ]
```
If this were to be released, in Express the function would be used like this:

```javascript
import express from 'express'

const app = express()

app.get('/', (req, res) => {
  res.send('Hello World')
})

console.log(app.router.getRoutes())
// [
//   {
//     name: 'handle',
//     path: '/',
//     methods: ['GET'],
//     keys: undefined,
//     router: undefined,
//     options: { strict: true, caseSensitive: true, end: true },
//   }
// ]
```

#### Why not reverse-engineer the path?

In Express 4, there were some hacks based on inspecting the generated regular expressions. But these approaches aren't solid or future-proof. If `path-to-regexp` changes the regex structure, those solutions break. So it's not a sustainable or effective approach.

Furthermore, there are plans to reintroduce some features to `path-to-regexp` (https://github.com/pillarjs/path-to-regexp/issues/380, https://github.com/pillarjs/path-to-regexp/issues/379), so it's highly likely that the internal regex structure will change again. On top of that, unlike in Express 4, the generated regex is no longer directly accessible.

close https://github.com/expressjs/express/discussions/5961, close https://github.com/expressjs/express/issues/6481, close https://github.com/pillarjs/router/issues/122).


jonchurch:
We need clarification on the exact requirements. Can you have them open a feature request?

There are different things that could be exposed, which of these is the request?

### 1. what routes did a user register?
  * e.g. `router.get('/users/:id')` returns `{ path: '/users/:id', methods: ['GET'] }`
  * This is what most of the linked issues want, for documentation/debugging purposes
  * This is what the current implementation in PR provides

### 2. what will match what the user registered?
* e.g. with `strict: false` (default) `/users` matches `/users` and `/users/`, with `strict: true` it does not
* e.g. with `caseSensitive: false` (default) `/Users` matches `/Users` and `/users`, with `caseSensitive: true` it does not
* This is necessary for **traffic routing purposes**, and what Im assuming vercel wants to do, route paths into the express app without passing through all traffic (assuming here without more info)
* The implementation does not cover this

If the ask is number 2, they would need to decide if they apply the routing rules themselves, tell users which options are supported, or we'd need to enhance the ouput to include either router options or otherwise take options into account.

(v good PR description ❤️)

jonchurch (changes requested):
https://github.com/pillarjs/router/pull/174#issuecomment-3129927730

test/mapRoutes.js
```diff
@@ -0,0 +1,115 @@
...
+        { path: '/', methods: ['_ALL'] },
+        { path: '/test2/', methods: [] },
+        { path: '/test/', methods: ['GET'] },
+        { path: '/^\\/[a-z]oo$/', methods: ['_ALL'] },
+        { path: '/foo', methods: ['GET'] },
+        { path: '/bar', methods: ['GET'] },
+        { path: '/:id/setting/:thing', methods: ['POST'] }
+      ])
+  })
+
```
jonchurch, line 35:
```suggestion
 it('should not create double slashes when mounting at root /', function () {
    const router = new Router()
    const subRouter = new Router()

    subRouter.get('/api', () => {})

    router.use('/', subRouter)

    const routes = router._mapRoutes()

    assert.deepStrictEqual(routes, [
      { path: '/api', methods: ['GET'] }
    ])
  })

```

jonchurch: fails currently due to the path concat in `collectRoutes`

index.js
```diff
@@ -450,6 +471,83 @@ methods.concat('all').forEach(function (method) {
...
+ * @private
+ */
+function collectRoutes (stack, prefix, routeMap) {
+  for (const layer of stack) {
+    // for routes without a .use
+    if (layer.pathPatterns && layer.route) {
+      const methods = Object.keys(layer.route.methods).map((method) => method.toUpperCase())
+      if (Array.isArray(layer.pathPatterns)) {
+        for (const pathPattern of layer.pathPatterns) {
+          const fullPath = prefix + pathPattern
```
jonchurch, line 527:
```suggestion
          const fullPath = prefix === '/' ? pathPattern : prefix + pathPattern
```
to fix https://github.com/pillarjs/router/pull/174/files#r2237920545

index.js
```diff
@@ -441,6 +441,27 @@ Router.prototype.route = function route (path) {
   return route
 }
 
+/**
+ * List all registered routes.
+ *
+ * @return {Array} An array of route paths
+ * @private
```
jonchurch, line 448:
```suggestion
 * @public
```
This is on the proto so it's public

bjohansebas:
@jonchurch thanks for the quick review! Yes, what they shared with me (and I also asked them to comment here about their needs) is that they want to be able to see the routing structure and then map it to their Build Output API (https://vercel.com/docs/build-output-api).

In my opinion, the solution you're proposing (combining both outputs) is also the right one, as it covers those needs as well as other use cases or certain hosting providers. Plus, it's flexible enough for developers to adapt it to their specific requirements.

```javascript
[
    {
      path: '/api',
      methods: ['GET'],
      options: { strict: false, caseSensitive: false }
    },
    {
      path: '/v1/users',
      methods: ['GET'],
      options: { strict: true, caseSensitive: true }
    }
  ]
``` ref https://github.com/pillarjs/router/pull/174#issuecomment-3129927730

File: index.js (hunk 2 above)
Comment (bjohansebas) [reply index.js:527]: thanks, good catch

File: index.js (hunk 3 above)
Comment (bjohansebas) [reply index.js:448]: Yep, by "private" I meant the underscore in the function name, but I had some doubts about doing it that way. It's also true that anyone could still access it, so it doesn't really prevent usage. Either way, I'll remove the underscore from the function name and document it as public. Any suggestions for the function name are more than welcome!

File: index.js [hunk 4]
```diff
@@ -441,6 +442,33 @@ Router.prototype.route = function route (path) {
   return route
 }
 
+/**
+ * List all registered routes.
+ *
+ * @return {Array} An array of route paths
+ * @public
+ */
+Router.prototype.mapRoutes = function mapRoutes () {
```
wesleytodd, line 451: `map` implies you can give it a function that will iterate over the routes. Since that is note the case, I would prefer one of two things:

1. `.routes()`
2. `get routes`

index.js
```diff
@@ -441,6 +442,33 @@ Router.prototype.route = function route (path) {
...
+  const routes = []
+  const stack = this.stack
+
+  const options = {
+    strict: this.strict,
+    caseSensitive: this.caseSensitive
+  }
+
+  const routeMap = new Map()
```
wesleytodd, line 460: Not sure I understand why this is converted from a `Map` back to an array. If we are going with a `Map` I would rather just return the `Map` and let the consumer of this do what they need with it.

index.js
```diff
@@ -441,6 +442,33 @@ Router.prototype.route = function route (path) {
...
+
+  collectRoutes(stack, '', routeMap, options)
```
wesleytodd, line 462: I understand why this is being done this way, but I am always concerned when using a function with side effects like this. I think we should consider having this public router method take an optional input for the `Map` and then instead of recursing with `collectRoutes` just use `router.routes()` (assuming we go with that naming I suggested above).

index.js
```diff
@@ -450,6 +478,101 @@ methods.concat('all').forEach(function (method) {
...
+ * Add a route to the map with the given path and methods.
+ * @param {Map} routeMap
+ * @param {string} path
+ * @param {Array} methods
+ * @private
+ */
+function addRouteToMap (routeMap, path, methods, options) {
+  if (routeMap.has(path)) {
+    const existingMethods = routeMap.get(path)
+    for (const method of methods) {
```
wesleytodd, line 491: I will be suggesting we keep more than just the `path` to `method` in here, and with that in mind I was thinking we may want to do something that would also avoid having to iterate `methods` like is being done here.

index.js
```diff
@@ -450,6 +478,101 @@ methods.concat('all').forEach(function (method) {
...
+        existingMethods.methods.push(method)
+      }
+    }
+  } else {
+    const { keys } = pathRegexp.pathToRegexp(path)
+
+    routeMap.set(
+      path,
+      {
+        methods: [...methods],
```
wesleytodd, line 502: Can this not just be `methods,`?

test/mapRoutes.js
```diff
@@ -0,0 +1,189 @@
...
+    router.route('/test2/')
+    router.route('/test/').get(noop)
+    // With regex patterns path-to-regexp fail
+    // router.all(/^\/[a-z]oo$/, noop)
+    router.get(['/foo', '/bar'], noop)
+    router.post('/:id/setting/:thing', noop)
+
+    assert.deepStrictEqual(router.mapRoutes(),
+      [
+        { path: '/', methods: ['_ALL'], keys: [], options: { strict: undefined, caseSensitive: undefined } },
```
wesleytodd, line 27: Ah, I didn't notice this magic sting above, but I am not a fan of `_ALL`. We should list them here.

wesleytodd (changes requested):
I left a few comments, but I think we should workshop this just a bit. I think ideally we don't need to do this work when we call `routes()` but instead we can keep internal track of these in such a way that they are much more easily operated on. That would likely only be possible in the next major, but if we can design this api such that it is our "ideal internal book keeping structure" then we can ease that transition in the next breaking change.

So I think my ask is that we add some usage examples in the main text of the PR (or maybe in the readme as docs in this pr) and then we can propose some small changes to the exposed api that way before we invest too much more work in the internals and tests?

EDIT: also, I was thinking that maybe one good pressure test for the api would be if I tried to integrate it into my openapi generator package. If it works well for that, I think it would work well for all the similar use cases. If we can organize this work I am happy to make it a priority.

index.js
ctcpip: +1 for `getRoutes`

bjohansebas:
> I have thoughts on this PR, but won't have time to review for a bit. If it's urgently needed, I don't want to be the blocker, but otherwise would like time to review. 🙏

It's not entirely urgent, so feel free to do the review whenever you can.

blakeembrey (commented):
This looks like a great starting point, love it! @jonchurch has some really good feedback to think about here too, re what Vercel may need.

Can Vercel handle the regex support?

index.js
```diff
@@ -450,6 +478,101 @@ methods.concat('all').forEach(function (method) {
...
+          addRouteToMap(routeMap, fullPath, methods, options)
+        }
+      } else {
+        const fullPath = prefix === '/' ? layer.pathPatterns : prefix + layer.pathPatterns
+        addRouteToMap(routeMap, fullPath, methods, options)
+      }
+    }
+
+    // for layers with a .use (mounted routers)
+    if (layer.pathPatterns && layer.handle && layer.handle.stack && !layer.route) {
```
blakeembrey, line 554: For `.use`, could it make sense to allow consumers to iterate recursively themselves? It's something that could be added in a follow up, but the MVP could be simply `{ path, method, router }`. Every field could also be optional, I guess, since `path: undefined` with `.use(fn)`, `method: undefined` when `all` is used, and `router: undefined` when no nested router.

index.js
```diff
@@ -450,6 +478,101 @@ methods.concat('all').forEach(function (method) {
...
+function addRouteToMap (routeMap, path, methods, options) {
+  if (routeMap.has(path)) {
+    const existingMethods = routeMap.get(path)
+    for (const method of methods) {
+      if (!existingMethods.methods.includes(method)) {
+        existingMethods.methods.push(method)
+      }
+    }
+  } else {
+    const { keys } = pathRegexp.pathToRegexp(path)
```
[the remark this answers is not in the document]
bjohansebas: that was something @wesleytodd asked for, exposing the keys
https://openjs-foundation.slack.com/archives/C02QB1731FH/p1753730868873529

index.js
```diff
@@ -450,6 +478,101 @@ methods.concat('all').forEach(function (method) {
...
+ * @param {string} path
+ * @param {Array} methods
+ * @private
+ */
+function addRouteToMap (routeMap, path, methods, options) {
+  if (routeMap.has(path)) {
+    const existingMethods = routeMap.get(path)
+    for (const method of methods) {
+      if (!existingMethods.methods.includes(method)) {
+        existingMethods.methods.push(method)
```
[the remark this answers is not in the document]
bjohansebas: The path will never be undefined .use always sets the path to '/' when no path is explicitly provided in the arguments.

index.js
bjohansebas: I'm no longer using map, since now repeated routes will be returned :)

index.js
bjohansebas: Now that it no longer iterates over the methods, if there are duplicate routes, each method will go in its own item in the array instead of searching for the route to check the methods and append them.

index.js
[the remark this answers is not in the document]
bjohansebas: Cool, I’ll make that change soon, i’m currently looking into how to avoid combining the string with the regexp so that I can make that change and allow the use of regex again for this method.

index.js
bjohansebas: I've renamed it to getRoutes

IamLizu (changes requested):
Mostly looks good to me now. However, I wonder if there's possibility of infinite recurrsion at `collectRoutes`. 

Imagine a user acceddientally creates circular route references and then calls to `getRoutes` would go on forever.

Perhaps there should a protection, maybe check if the current layer is visited in the current branch. Here's a possible sample code,

```js
function collectRoutes (stack, prefix, routeMap, options, visited = new WeakSet()) {
  for (const layer of stack) {
    if (layer.pathPatterns && layer.handle && layer.handle.stack && !layer.route) {
      // Check for circular reference
      if (visited.has(layer.handle)) {
        continue  // Skip circular reference
      }
      
      visited.add(layer.handle)
      collectRoutes(layer.handle.stack, pathPrefix, routeMap, options, visited)
      visited.delete(layer.handle)  // Allow re-visiting in different branches
    }
  }
}
```

Other than this, I think `getRoutes` is awesome feature to the router.

Good work @bjohansebas 👏 

cc: @blakeembrey @wesleytodd @jonchurch 

#### Update

Just to be safe, @bjohansebas can we add more complex and nested routes in the test often found in real world scenarios to be safe?

index.js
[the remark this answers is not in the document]
bjohansebas: Yep, I’m referring to nested routes, for example:

```javascript
const router = new Router()
const inner = new Router()

inner.all(/^\/[a-z]oo$/, noop)
router.use("/t1", inner)

console.log(router.getRoutes())
// [
//   { path: "/t1/^\\/[a-z]oo$/", methods: ['_ALL'], keys: [], options: { strict: undefined, // caseSensitive: undefined }
// ]
``` 

With that route, I can no longer get the regex and extract the keys, so I think the best thing would be to stick with what [you suggested](https://github.com/pillarjs/router/pull/174#discussion_r2240468683). If I understand the idea correctly, the way 


```javascript
const router = new Router()
const inner = new Router()
const subinner = new Router

subinner.get("/t4")
inner.use("/t3", subinner)
inner.all(/^\/[a-z]oo$/, noop)
router.use("/t1", inner)
router.get("/t2, noop)

console.log(router.getRoutes())
[
 { path: "/t1", 
   keys: [], 
   methods: undefined, 
   options: {...},
   router: [
     { 
       path: "/t3", 
       keys: [], 
       methods: undefined, 
       router: [
         {
           path:"/t4", 
           methods: ["GET"], 
           keys: [], 
           router: undefined,
           options: {...}
         }
       ],
       options: {...}
     },
     {
       path: /^\/[a-z]oo$/, // Now I can get the keys from the regex
       keys: [],
       methods: ["_ALL"],
       router: undefined,
       options: {...}
     }
   ] 
 },
 {
   path: "/t2",
   keys: [],
   methods: ["GET"],
   router: undefined
   options: {...}
 }
]
```

blakeembrey:
Note: options will need to contain `end: boolean` for `.use` routes, which are prefix matched vs fully matched.

nronnei:
@bjohansebas you're my hero; eagerly awaiting this change.

Megapixel99:
Thank you for your work on this pull request. While I was working on: https://github.com/wesleytodd/express-openapi/pull/77, I had to rely on the changes in this PR to get the full path of the routes. 

Would it be possible to expose the middleware stack of each route within `getRoutes()`? Currently I have to get the middleware stack directly from the router (https://github.com/wesleytodd/express-openapi/pull/77/files#diff-ffcbf454c1de79602c5a1cbbc810c181b8d8552f743952d9049ffc13ca8b1f1eR31-R33) despite using `getRoutes()` to get the full path of the URL earlier in my code (https://github.com/wesleytodd/express-openapi/pull/77/files#diff-ffcbf454c1de79602c5a1cbbc810c181b8d8552f743952d9049ffc13ca8b1f1eR12)

bjohansebas:
Yep, that could be added in my opinion. I need to partially modify this PR to adapt it to the idea in [https://github.com/pillarjs/router/pull/174#discussion\_r2246484011](https://github.com/pillarjs/router/pull/174#discussion_r2246484011), I have it in my priorities, so you’ll need to adapt your PR to that format once I make the change.

jeffsee55:
👋 we (Vercel) have just added an experimental feature in our Express builder that allows us to extract metadata from the Express app like routes, static assets folders, and view rendering engines so we can build Express apps with first-class features like o11y out of the box. 

[Here](https://github.com/vercel/vercel/pull/13963)'s the initial PR with more details, but at the moment the route metadata is very basic (doesn't capture sub-routers) so I wanted to check in on this PR and ask if there's anything we can do to make route introspection better. 

You can try it out by using [this repo](https://github.com/jeffsee55/express-with-o11y/), there's a Vercel deploy button in the README.

tknickman:
@bjohansebas - Anything we can help with to get this shipped? We're excited to leverage this on Vercel's side.

bjohansebas:
Hey, sorry for the delay, i’ve been away from open source for a while, and I’m slowly getting back to catching up on all the pending work.

@tknickman and express community, it would be great to know if you have any thoughts on how this API should be presented. We can’t cover every possible route style that this library allows, but we can try to handle the majority of them.

What I still have pending to do is:
- Retrieve the keys when the route path is a regexp
- Address request https://github.com/pillarjs/router/pull/174#issuecomment-3237904000, which I think is a good option

IamLizu (changes requested):
Hey @bjohansebas!

I think there's infinite recursion on cyclic routers (index.js lines ~479-548). `collectRoutes` walks mounted routers without a visited-set, so any accidental cycle (reusing routers across mounts) makes `getRoutes()` blow the stack. For example,

   ```js
   const Router = require('./')

   const app = new Router()
   const admin = new Router()
   const account = new Router()

   admin.get('/dashboard', () => {})
   account.get('/:accountId', () => {})

   app.use('/admin', admin)
   app.use('/accounts', account)

   // Later refactor reuses routers and creates a cycle:
   admin.use('/accounts', account)
   account.use('/admin', admin)

   app.getRoutes() // RangeError: Maximum call stack size exceeded
   ```

   Maybe add a `WeakSet` (or similar) guard plus a regression test that mounts routers cyclically?

blakeembrey:
@tknickman Does Vercel need the keys from strings and regexes? What's the MVP needed?

Asking because it feels like the MVP to ship could be an array with no recursion (exposing `router`, not nested routes), and skip keys for now. With that additional features can be exposed in the array in a follow up PR.

For the original PR use-case of a flattened list that could be easily consumed, that is the type of thing that can be added and exported from the main package once this API is stable, since it's just a more user-friendly (but specific) output.

jeffsee55:
@blakeembrey we just need the paths themselves, and ideally the methods. Eg. `/users/:id` and `[GET]`, etc

bjohansebas:
> Asking because it feels like the MVP to ship could be an array with no recursion (exposing router, not nested routes), and skip keys for now. With that additional features can be exposed in the array in a follow up PR.

For the context of the keys https://github.com/pillarjs/router/pull/174#discussion_r2240506353

blakeembrey:
@bjohansebas I get that, but it's something that can be added in another PR without a breaking change. It seems like there's the bones of an implementation that can be shipped today and iterated on without any breaking changes if you define the basic API clearly.

bjohansebas:
By the way, I’ve already requested internally to get this unblocked, since there are many different approaches.

README.md
```diff
@@ -143,6 +143,60 @@ router.param('user_id', function (req, res, next, id) {
 })
 ```
 
+### route.getRoutes()
```
jonchurch, line 146: Wont block on this but Id prefer to see `route.routes()` or `route.listRoutes()` to avoid intellisense/tab complete conflict when folks type `app/router.g` in an attempt to type `.get`

blakeembrey:
@bjohansebas We discussed this on the call today and @wesleytodd is on board with landing something simple (re https://github.com/pillarjs/router/pull/174#issuecomment-3577770290) and iterating from there. TL;DR is exposing just `{ methods, path, router }`, and then things like `keys` or `handler` can be added later upon request.

Additionally, exposing the `router` instead of the resolved `listRoutes` recursive call removes the concerns around infinite recursion and leave us a path to supporting other nested routers in the future.

The only other comment would be changing `_ALL` toward something else, e.g. `methods: []` or `methods: "*"` perhaps, whatever is easy to document. Is that a reasonable change?

@jonchurch Has a valid point around renaming it too, `router.routes()` or `router.listRoutes()`. 

Final question on exposing `router`, can we just expose `handler` and let people do `typeof handler.listRoutes === "function"` to work recursively?

blakeembrey:
@UlisesGascon This isn't a major, removing the tag.

README.md
wesleytodd: I think `listRoutes()` is reasonable.

index.js
```diff
@@ -441,6 +443,23 @@ Router.prototype.route = function route (path) {
...
+Router.prototype.getRoutes = function getRoutes () {
+  const stack = this.stack
+
+  const options = {
+    strict: this.strict,
+    caseSensitive: this.caseSensitive
+  }
+
+  return collectRoutes(stack, options)
+}
```
wesleytodd, line 461: Small formatting nitpick (not required, just personal preference):


```suggestion
Router.prototype.getRoutes = function getRoutes () {
  return collectRoutes(this.stack, {
    strict: this.strict,
    caseSensitive: this.caseSensitive
  })
}
```

lib/layer.js
```diff
@@ -43,7 +44,11 @@ function Layer (path, options, fn) {
   this.keys = []
   this.name = fn.name || '<anonymous>'
   this.params = undefined
+  // path is determinate in runtime execution
   this.path = undefined
+  this.end = opts.end
+
+  this.pathPatterns = path
```
wesleytodd, line 51: The naming of this implies to me two things:

1. it is always a list (which it is not, it can be a few things)
2. That it is related somehow to `URLPattern` (which it is not, but we would want to support someday so avoiding naming confusion now seems good)

Maybe a nitpick the naming would be we could call it `rawPath` or something similar to that so that it is clear this is "what the user passed in"?

index.js
wesleytodd: Agreed here, lets just simplify this down into the bare minimum api before merging this PR. Then we can take on nesting routers and deeply iterating the stack more directly later if we deem it worth it. We want to unstuck this progress, and I think these details is what I got hung up on last time leading to my original blocking review. Better to land what we can for sure agree on adds value and move this forward.

README.md
```diff
@@ -143,6 +143,60 @@ router.param('user_id', function (req, res, next, id) {
...
+//         router: undefined,
+//         options: { strict: true, caseSensitive: false, end: true },
+//       }
+//     ],
+//     options: { strict: true, caseSensitive: true, end: false }
+//   },
+//   {
+//     name: 'handle',
+//     path: '/:id',
+//     methods: ['_ALL'],
```
wesleytodd, line 192: We discussed on the call taking an alternative approach to `_ALL` and just using `undefined` to indicate "all supported methods". The problem is that almost *no* route middleware really supports "all methods", so this is almost always just the case for `.use` and would need to *know* what methods the other routes following after support.


```suggestion
//     methods: undefined,
```

wesleytodd (approved):
Explicitly removing my blocking review. We need to address @blakeembrey's feedback on simplification before landing, but with my limited availability I don't want to be the one blocking the merge.

Added a few nitpick comments, but once we simplify it down I am good to merge and publish.

fed135:
Hey all! I've been working on express plugins, often needing to monkey-patch internals to get all the routes in a router. I had a different approach to this problem: https://github.com/expressjs/express/pull/7321 

I feel that this PR here addresses it in a much cleaner fashion. I'd close mine as redundant but there's one part that is missing in the returned objects for `getRoutes`: the handler stack. This would be super useful to have, even as a Frozen/Immutable.

For reference, the object I'm emitting in my implementation is the plain Route object as-is.

```
Route {
    path: '/:foo',
    stack: [ [Layer] ],
    methods: [Object: null prototype] { get: true }
  }

``` 

Would there be openness to adding the `stack` property?

bjohansebas:
Okay, I had overlooked this. I've added it to my backlog of things to finish

bjohansebas:
Okay, here are the changes I've made so far and how the API behaves now.

 each `Layer` stores the path it was
registered with in `rawPath` for arrays, a `flat(Infinity)` that both
copies (immune to later mutations of the caller's array) and unnests. This
is the only cost added to the hot path: one assignment per layer.

1. **Route** (`layer.route` is set) → `methods = route._methods()`: the
   registered methods in uppercase, including the automatic `HEAD` for
   `GET` routes, with the `_all` sentinel filtered out by
   `Route._methods()` itself. If the result is empty and the route was
   registered only with `.all()` → `undefined` ("matches all methods");
   if it is empty without `.all()` → `[]` (a route with no handlers).
2. **Mounted router** (`Array.isArray(handle.stack)`) → an entry with
   `router: handle` and `methods: undefined`. No recursion: consumers
   recurse themselves by calling `router.listRoutes()` when available —
   which eliminates the cyclic-router problem by design, and the
   shape-based detection also lists routers from older versions of this
   package.
3. **Plain middleware** → skipped.

**Output**: one `{ path, methods, router }` entry per path (array paths
are expanded, each with its own copy of `methods`)

```js
router.get('/test', h1)
router.get('/test', h2)
router.post('/test', h3)

router.listRoutes()
// [
//   { path: '/test', methods: ['GET', 'HEAD'], router: undefined },
//   { path: '/test', methods: ['GET', 'HEAD'], router: undefined },
//   { path: '/test', methods: ['POST'], router: undefined }
// ]
```

PR #178, opened 2025-09-12
feat: Prioritize strict matches over paths with parameters

asynchroza:
We're using NestJS in a new project which depends on Express and the router package.

We have multiple paths structured in the following way: `/controls/:uuid`, `/controls/executions`. These two paths are part of separate `Controller` NestJS classes which means that we don't have much control over the order.

The issue is that when `router` is trying to match the path, the `/controls/:uuid` comes first and since it's expecting a string as the path parameter, the regex returns true when the requested path is `/controls/executions`.

This PR introduces a fix for this by extending the search for other routes which match the pattern instead of exiting immediately on a match.


asynchroza:
I do not know whom to ask but I guess I will direct my question towards you - @wesleytodd , @dougwilson. Would the team be interested in something similar? This particular implementation seems to fail some of the tests and since I'm not that familiar with the structure of the project yet, I figured this is a good starting point. If provided with input, I'm willing to work on the approach and fix it.

asynchroza:
I've tested this implementation in our project and seems to have similar benchmarks but since it's failing a lot of tests, we will not be deploying to production and instead restructure our endpoints.

asynchroza:
We've found a fix for our project - reorder the multiple different controllers in the module's list which is then used for initialization.

PR #183, opened 2025-10-07
feat: add router cycle detection to prevent infinite loops

aprici7y:
Detects when a router is mounted on itself (directly or indirectly) and emits a warning to help developers identify cycles that cause requests to hang indefinitely.

When a router is mounted on itself through a chain of routers, requests that traverse the cycle hang indefinitely and eventually cause a stack overflow. This is difficult to debug because the application appears to work fine until a request hits the cyclic path.

This implements cycle detection in Router.prototype.use that:
- Detects both direct self-mounts (router.use(router)) and indirect cycles (routerA -> routerB -> routerA)
- Uses recursive traversal with a visited set to prevent infinite loops during detection
- Emits a process.emitWarning() with type RouterCycleWarning
- Maintains backward compatibility - the mount still succeeds but with a warning

## Manual Test

```javascript
const Router = require('router')

// Direct self-mount
const loop = new Router()
loop.use('/', loop) // Emits RouterCycleWarning

// Indirect cycle
const routerA = new Router()
const routerB = new Router()
routerA.use('/b', routerB)
routerB.use('/a', routerA) // Emits RouterCycleWarning
```

Fixes https://github.com/expressjs/express/issues/6809


index.js
```diff
@@ -409,6 +420,51 @@ Router.prototype.use = function use (handler) {
...
+ * Detect if mounting a router would create a cycle.
+ *
+ * @param {Router} parentRouter - The router being mounted onto
+ * @param {Router} childRouter - The router being mounted
+ * @param {Set} visited - Set of visited routers to avoid infinite recursion
+ * @return {boolean}
+ * @private
+ */
+
+function detectRouterCycle (parentRouter, childRouter, visited) {
```
IlyasShabi, line 433: You run this function each time, so why not adding a Set/Map to avoid repetitive code?

som14062005:
I've opened a fresh PR addressing the same issue with a cleaner implementation: #193

## Implementation Details

The fix adds a `hasCycle()` helper function that performs a depth-first 
search (DFS) through the router tree before mounting any router instance.

- In `Router.prototype.use`, before pushing a layer onto the stack, 
  we check if the handler is a `Router` instance
- If it is, `hasCycle()` walks the entire subtree of that router using 
  a `visited` Set to avoid infinite loops during the check itself
- If a cycle is detected, `process.emitWarning()` is called with code 
  `ROUTER_CYCLE_DETECTED`, giving developers an immediate, actionable warning
- If no cycle exists, everything proceeds as normal — zero performance 
  impact on the happy path

PR #190, opened 2026-02-13
fix: prevent `use` with multi-slash path from matching all routes

MOHIT51196:
## Problem
`router.use('//', handler)` incorrectly matches **every** incoming request, including `/` and `/foo`. 
A sub-router mounted at `//` should only respond to requests whose path actually begins with `//`.
Reported upstream in expressjs/express#4557. Also related to expressjs/express#3798.

### Reproduction
```
const router = Router();
const inner = Router();

inner.get('/', listTestRoutes);
inner.get('/drop/table/:table', dropTable);

router.use('//', inner);
router.listen();

```

**Expected**: GET / does not match the inner router.
**Actual**: GET / matches inner.get('/') and serves the test route list.

### Root cause
The loosen() helper in lib/layer.js strips trailing slashes from mount paths using the regex /\/+$/ so that non-strict matching works (e.g. /foo/ loosens to /foo).
For a path consisting entirely of slashes like //:
`String('//').replace(/\/+$/, '')  →  ''  (empty string)`

An empty string passed to path-to-regexp.match('', { end: false }) creates a pattern that matches every path, because every path starts with an empty prefix. This is the bug.

### Fix
A one-line guard in loosen(): if stripping trailing slashes produces an empty string, preserve the original path instead.
```
const loose = str.replace(TRAILING_SLASH_REGEXP, '')
return loose || str
```
This ensures:

- /foo/ still loosens to /foo (normal behavior, unchanged)
- // is preserved as // (no longer collapses to empty string)
- /// is preserved as ///

### Changes

- **lib/layer.js** — Guard loosen() against reducing slash-only paths to empty string
- **test/router.js** — 3 new test cases:
  - use('//') must not match / or /foo
  - use('///') must not match / or /foo
  - Sub-router mounted at // must not expose its routes via single /

>All 697 tests pass (694 existing + 3 new). Linter clean.
> Fixes [expressjs/express#4557](https://github.com/expressjs/express/issues/4557)


MOHIT51196:
@wesleytodd can you review this PR with a fix for an open issue expressjs/express#4557

mohitmalhotra-celigo:
@UlisesGascon can you review this PR with a fix for an open issue expressjs/express#4557

PR #193, opened 2026-03-31
fix: detect and warn on circular router references

som14062005:
## Problem
When a Router is mounted within itself (directly or transitively),
requests that traverse the cycle hang indefinitely with no error or warning.

## Solution
- Added `hasCycle()` helper using DFS with a `visited` Set to avoid infinite loops
- In `Router.prototype.use`, checks for circular references before mounting a Router instance
- If a cycle is detected, emits a process warning with code `ROUTER_CYCLE_DETECTED`

## Notes
- 11 pre-existing test failures related to the QUERY method exist on the
  base branch and are unrelated to this PR
- Zero new test failures introduced by this change


som14062005:
## Implementation Details

The fix adds a `hasCycle()` helper function that performs a depth-first 
search (DFS) through the router tree before mounting any router instance.

- In `Router.prototype.use`, before pushing a layer onto the stack, 
  we check if the handler is a `Router` instance
- If it is, `hasCycle()` walks the entire subtree of that router using 
  a `visited` Set to avoid infinite loops during the check itself
- If a cycle is detected, `process.emitWarning()` is called with code 
  `ROUTER_CYCLE_DETECTED`, giving developers an immediate, actionable warning
- If no cycle exists, everything proceeds as normal — zero performance 
  impact on the happy path

PR #196, opened 2026-04-15
feat: add request layer lifecycle TracingChannels

logaretm:
This PR adds a single [tracing channel](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) that emits structured lifecycle events (start, end, asyncStart, asyncEnd, error) for every middleware, route handler, and error handler execution.

This builds on #96 but uses tracing channels instead which enables context propagation out of the box.

### Context shape

The context object is pretty minimal, providing both `req` and `res` enables APMs to figure out how to react to that specific payload.

```js
{ req, res, layer }
```

The `layer` is the Layer instance. Subscribers use `layer.name`, `layer.handle.length` (4 for error handlers), and `req.route` (set when inside a matched route) to classify and name spans however they see fit.

### Usage Example:

This is an example I've set up locally using a Sentry setup with this PR patched in:

```ts
const Sentry = require('@sentry/node')
// otel-tracing-channel is a simple tracing channel wrapper that patches in the OTEL context propagation
// atm otel doesn't support tracing channels, so this workaround is required for OTEL-based APMs
// the user wouldn't need this and typically they can set their own thing with ALS.
const { tracingChannel } = require('otel-tracing-channel')

Sentry.init({
  // ...
})

// Subscribe to express:request TracingChannel and create Sentry spans
const channel = tracingChannel('pillarjs.router.request', (ctx) => {
  const { layer, req } = ctx
  const isErrorHandler = layer.handle.length === 4
  const isRouteHandler = Boolean(req.route)

  const spanName = isRouteHandler
    ? `${req.method} ${req.route.path}`
    : isErrorHandler
      ? `error_handler - ${layer.name}`
      : `middleware - ${layer.name}`

  return Sentry.startSpanManual(
    {
      name: spanName,
      op: isRouteHandler ? 'http.handler' : isErrorHandler ? 'middleware.error' : 'middleware',
      attributes: {
        'express.name': layer.name,
        'http.request.method': req.method,
        'url.path': req.url,
        'url.full': req.originalUrl,
        ...(req.route && { 'http.route': req.route.path }),
        ...(req.query && Object.keys(req.query).length > 0 && {
          'url.query': JSON.stringify(req.query)
        }),
        ...(req.params && Object.keys(req.params).length > 0 && {
          'express.params': JSON.stringify(req.params)
        })
      }
    },
    (span) => span
  )
})

channel.subscribe({
  asyncEnd (ctx) {
    if (ctx.req.route) {
      ctx.span?.setAttribute('http.response.status_code', ctx.res.statusCode)
    }
    ctx.span?.end()
  },
  error (ctx) {
    ctx.span?.setStatus({ code: 2, message: ctx.error?.message })
    ctx.span?.end()
  }
})

// Express App
const express = require('express')
const app = express()

// Middleware
app.use(express.json())

app.use(function requestLogger (req, res, next) {
  console.log(`${req.method} ${req.url}`)
  next()
})

app.use(function addRequestId (req, res, next) {
  req.requestId = Math.random().toString(36).slice(2, 10)
  res.setHeader('x-request-id', req.requestId)
  next()
})

// Routes
app.get('/users/:id', function validateId (req, res, next) {
  if (isNaN(req.params.id)) {
    return res.status(400).json({ error: 'id must be a number' })
  }
  next()
}, function loadUser (req, res, next) {
  req.user = { id: req.params.id, name: 'Test User' }
  next()
}, function getUser (req, res) {
  res.json({ ...req.user, requestId: req.requestId })
})
```

This produces a waterfall view looking like this:

CleanShot 2026-04-15 at 08 36 22@2x

Note: The three spans that look the same at the bottom actually represent the three handlers executing for that route, it is up to the subscriber to decide which one is the route handler and how to name the other ones since it has access to `layer`.

### Is tracing channels experimental?

We had this come up a few times during the initiative of pushing the adoption of dc across the ecosystem, most recently this was brought up and responded to by @Qard here:

https://github.com/Automattic/mongoose/issues/16105#issuecomment-4338019446

> Yes, it has mainly been left as experimental due to a lack of adoption, but also as a signal that we may make improvements as we learn from uses of it. That learning has largely already been absorbed and applied via recent PRs at this point. There are several improvements incoming, but none of those change the existing code, which will be left as-is, they add some additional systems to handle some more specific cases such as async iterators.

> The TracingChannel API will be left as-is, likely forever at this point as it is very important to us to retain compatibility essentially indefinitely, because APM vendors have many customers using old Node.js versions even well beyond the EOL of Node.js itself [...]

### Related work

- This builds on #96 but uses tracing channels
- Related to https://github.com/expressjs/express/pull/7041
- Does the same as https://github.com/expressjs/express/pull/7171 but uses tracing channels
- Closes https://github.com/expressjs/express/issues/6353

### Caveats

- The channel name right now is `pillarjs.router.request`, in the original PR it was `request` but I think channel names need to have a relationship with the module name otherwise we risk duplication in the ecosystem.
- Depending on the exact version of Node 18 certain behaviors are expected in this PR:
  - Node versions without `hasSubscribers` will assume that there is a subscriber and will execute the traced path (with context object allocation). Multiple libraries like `redis` and `mysql2` accepted this limitation.
  - Node versions without `tracingChannel` support will just not offer tracing of any kind. The fallback here is NOOP.
- It is up to APMs to represent the onion shape with nested handlers or flatten certain spans which is relatively easy if APMs have their own ALS context propagation strategy.
- Tests running a node version that do not have `tracingChannel` will be skipped via `describe.skip`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,6 +400,56 @@ router.route('/pet/:id')
 server.listen(8080)
 ```
 
+## Diagnostics
+
+`router` integrates with Node.js [`diagnostics_channel`](https://nodejs.org/api/diagnostics_channel.html)
+via a [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel)
+named `pillarjs.router.request`. This lets observability tools (APMs, tracers,
+loggers) hook into middleware and route handler execution without monkey-patching.
+
+Each layer's handler invocation publishes the standard tracing channel sub-events
+(`start`, `end`, `asyncStart`, `asyncEnd`, `error`). The published context object
+contains:
+
+- `req`: the incoming `http.IncomingMessage`
+- `res`: the `http.ServerResponse`
+- `layer`: the internal `Layer` instance being invoked (exposes `.name`, `.path`, `.handle`, etc.). Note that `Layer` is an internal implementation detail and its shape may change between releases.
+- `error`: the error passed to `next(err)`, when applicable
+- `handled`: `true` when the layer is an error-handling middleware (4-arg signature)
+
+The `error` event is also published when a handler calls `next(err)` with a real
+error. The control-flow sentinels `'route'` and `'router'` are not treated as
+errors and will not publish to the `error` channel.
+
+When no subscribers are attached, tracing is bypassed entirely, so there is no
+context allocation or channel publishing overhead on the hot path.
+
+```js
+const dc = require('node:diagnostics_channel')
+
+const channel = dc.tracingChannel('pillarjs.router.request')
+
+channel.subscribe({
+  start (ctx) {
+    ctx.startTime = process.hrtime.bigint()
+  },
+  end (ctx) {
+    // do whatever you need on synchronous completion
+  },
+  asyncStart (ctx) {
+    // do whatever you need when the async portion begins
+  },
+  asyncEnd (ctx) {
+    const durationNs = process.hrtime.bigint() - ctx.startTime
+    console.log('%s %s -> %s (%dns)',
+      ctx.req.method, ctx.req.url, ctx.layer.name, durationNs)
+  },
+  error (ctx) {
+    console.error('handler error in %s:', ctx.layer.name, ctx.error)
+  }
+})
+```
+
 ## License
 
 [MIT](LICENSE)
--- a/README.md
+++ b/README.md
@@ -404,7 +404,7 @@ server.listen(8080)
 
 `router` integrates with Node.js [`diagnostics_channel`](https://nodejs.org/api/diagnostics_channel.html)
 via a [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel)
-named `pillarjs.router.request`. This lets observability tools (APMs, tracers,
+named `express.router.request`. This lets observability tools (APMs, tracers,
 loggers) hook into middleware and route handler execution without monkey-patching.
 
 Each layer's handler invocation publishes the standard tracing channel sub-events
@@ -427,7 +427,7 @@ context allocation or channel publishing overhead on the hot path.
 ```js
 const dc = require('node:diagnostics_channel')
 
-const channel = dc.tracingChannel('pillarjs.router.request')
+const channel = dc.tracingChannel('express.router.request')
 
 channel.subscribe({
   start (ctx) {
--- a/README.md
+++ b/README.md
@@ -415,7 +415,7 @@ contains:
 - `res`: the `http.ServerResponse`
 - `layer`: the internal `Layer` instance being invoked (exposes `.name`, `.path`, `.handle`, etc.). Note that `Layer` is an internal implementation detail and its shape may change between releases.
 - `error`: the error the layer failed with, when applicable
-- `handled`: `true` when the layer is an error-handling middleware (4-arg signature)
+- `errorHandler`: `true` when the layer is an error-handling middleware (4-arg signature)
 
 The `error` event is published once, on the layer where the error originates,
 whether the handler calls `next(err)`, throws, or returns a rejected promise. An
--- a/README.md
+++ b/README.md
@@ -414,12 +414,14 @@ contains:
 - `req`: the incoming `http.IncomingMessage`
 - `res`: the `http.ServerResponse`
 - `layer`: the internal `Layer` instance being invoked (exposes `.name`, `.path`, `.handle`, etc.). Note that `Layer` is an internal implementation detail and its shape may change between releases.
-- `error`: the error passed to `next(err)`, when applicable
+- `error`: the error the layer failed with, when applicable
 - `handled`: `true` when the layer is an error-handling middleware (4-arg signature)
 
-The `error` event is also published when a handler calls `next(err)` with a real
-error. The control-flow sentinels `'route'` and `'router'` are not treated as
-errors and will not publish to the `error` channel.
+The `error` event is published once, on the layer where the error originates,
+whether the handler calls `next(err)`, throws, or returns a rejected promise. An
+error bubbling up through outer layers is not reported again. The `'route'` and
+`'router'` routing signals are not treated as errors and will not publish to the
+`error` channel.
 
 When no subscribers are attached, tracing is bypassed entirely, so there is no
 context allocation or channel publishing overhead on the hot path.
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -218,6 +218,12 @@ function recordError (req, err) {
   return true
 }
 
+// 'route' and 'router' signal to exit the route / router (see route.js), not
+// real errors.
+function isRoutingSignal (err) {
+  return err === 'route' || err === 'router'
+}
+
 /**
  * Invoke a handler wrapped in the request TracingChannel. Only called when the
  * channel has subscribers, so ctx is always present.
@@ -226,10 +232,9 @@ function recordError (req, err) {
 
 function invokeWithTrace (exec, ctx, next) {
   const wrappedNext = function (err) {
-    // 'route' and 'router' are control-flow sentinels (skip route / exit router),
-    // not real errors — don't pollute the error channel with them. An error that
-    // already bubbled up from an inner layer is only reported at its origin.
-    if (err && err !== 'route' && err !== 'router' && recordError(ctx.req, err)) {
+    // Don't pollute the error channel with routing signals, and only report an
+    // error at its origin, not again as it bubbles up through outer layers.
+    if (err && !isRoutingSignal(err) && recordError(ctx.req, err)) {
       ctx.error = err
       requestChannel.error.publish(ctx)
     }
@@ -237,7 +242,30 @@ function invokeWithTrace (exec, ctx, next) {
   }
 
   try {
-    const ret = requestChannel.tracePromise(function () { return handlePromise(exec(wrappedNext)) }, ctx)
+    const ret = requestChannel.tracePromise(function () {
+      // Catch thrown/rejected routing signals here so tracePromise doesn't
+      // report them as errors; real errors propagate and are reported as usual.
+      let out
+      try {
+        out = handlePromise(exec(wrappedNext))
+      } catch (err) {
+        if (isRoutingSignal(err)) {
+          next(err)
+          return
+        }
+        throw err
+      }
+
+      if (out) {
+        return out.then(undefined, function (err) {
+          if (isRoutingSignal(err)) {
+            next(err)
+            return
+          }
+          throw err
+        })
+      }
+    }, ctx)
 
     if (ret) {
       ret.then(null, function (error) {
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -33,11 +33,15 @@ const MATCHING_GROUP_REGEXP = /\((?:\?<(.*?)>)?(?!\?)/g
 
 const requestChannel = dc.tracingChannel && dc.tracingChannel('express.router.request')
 
-/**
- * Check if the channel has subscribers.
- */
+// TracingChannel#hasSubscribers is undefined before Node 20.13/22, so check the sub-channels.
 function shouldTrace (ch) {
-  return ch && ch.hasSubscribers !== false
+  return Boolean(ch) && (
+    ch.start.hasSubscribers ||
+    ch.end.hasSubscribers ||
+    ch.asyncStart.hasSubscribers ||
+    ch.asyncEnd.hasSubscribers ||
+    ch.error.hasSubscribers
+  )
 }
 
 /**
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -241,10 +241,10 @@ function invokeWithTrace (exec, ctx, next) {
     next(err)
   }
 
-  // All outcomes flow through wrappedNext (never tracePromise's own error
-  // publishing), so signal filtering and dedup happen in one place and tracing
-  // never changes how the router routes.
-  requestChannel.tracePromise(function () {
+  // runStores fires `start` and sets the async context; the rest are published
+  // manually, so a synchronous layer emits only start/end. All outcomes flow
+  // through wrappedNext to keep signal filtering and dedup in one place.
+  requestChannel.start.runStores(ctx, function () {
     let out
     try {
       out = handlePromise(exec(wrappedNext))
@@ -252,17 +252,30 @@ function invokeWithTrace (exec, ctx, next) {
       // A sync throw is forwarded verbatim, so a falsy value keeps routing
       // exactly as the untraced path does.
       wrappedNext(err)
+      requestChannel.end.publish(ctx)
       return
     }
 
-    if (out) {
-      return out.then(undefined, function (err) {
+    if (!out) {
+      requestChannel.end.publish(ctx)
+      return
+    }
+
+    requestChannel.end.publish(ctx)
+    out.then(
+      function () {
+        requestChannel.asyncStart.publish(ctx)
+        requestChannel.asyncEnd.publish(ctx)
+      },
+      function (err) {
         // A rejected promise is an error even when the reason is falsy;
         // normalize it to the error the router forwards to next().
         wrappedNext(err || new Error('Rejected promise'))
-      })
-    }
-  }, ctx)
+        requestChannel.asyncStart.publish(ctx)
+        requestChannel.asyncEnd.publish(ctx)
+      }
+    )
+  })
 }
 
 /**
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -31,7 +31,7 @@ const MATCHING_GROUP_REGEXP = /\((?:\?<(.*?)>)?(?!\?)/g
  * @private
  */
 
-const requestChannel = dc.tracingChannel && dc.tracingChannel('express:request')
+const requestChannel = dc.tracingChannel && dc.tracingChannel('pillarjs.router.request')
 
 /**
  * Check if the channel has subscribers.
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -154,7 +154,7 @@ Layer.prototype.handleError = function handleError (error, req, res, next) {
   const layer = this
   invokeWithTrace(function (wrappedNext) {
     return fn(error, req, res, wrappedNext)
-  }, { req, res, layer, error, handled: true }, next)
+  }, { req, res, layer, error, errorHandler: true }, next)
 }
 
 /**
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -37,7 +37,7 @@ const requestChannel = dc.tracingChannel && dc.tracingChannel('express:request')
  * Check if the channel has subscribers.
  */
 function shouldTrace (ch) {
-  return ch && ch.start.hasSubscribers !== false
+  return ch && ch.hasSubscribers !== false
 }
 
 /**
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -180,7 +180,9 @@ function invokeWithTrace (exec, ctxFactory, next) {
   const ctx = tracing ? ctxFactory() : null
   const wrappedNext = tracing
     ? function (err) {
-      if (err) {
+      // 'route' and 'router' are control-flow sentinels (skip route / exit router),
+      // not real errors — don't pollute the error channel with them.
+      if (err && err !== 'route' && err !== 'router') {
         ctx.error = err
         // Explicitly publish the error to the error channel
         requestChannel.error.publish(ctx)
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -31,7 +31,7 @@ const MATCHING_GROUP_REGEXP = /\((?:\?<(.*?)>)?(?!\?)/g
  * @private
  */
 
-const requestChannel = dc.tracingChannel('express:request')
+const requestChannel = dc.tracingChannel && dc.tracingChannel('express:request')
 
 /**
  * Check if the channel has subscribers.
--- a/lib/layer.js
+++ b/lib/layer.js
@@ -232,8 +232,8 @@ function isRoutingSignal (err) {
 
 function invokeWithTrace (exec, ctx, next) {
   const wrappedNext = function (err) {
-    // Don't pollute the error channel with routing signals, and only report an
-    // error at its origin, not again as it bubbles up through outer layers.
+    // Routing signals aren't errors; only report a real error once, at its
+    // origin, not again as it bubbles up through outer layers.
     if (err && !isRoutingSignal(err) && recordError(ctx.req, err)) {
       ctx.error = err
       requestChannel.error.publish(ctx)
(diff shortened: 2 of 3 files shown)


test/tracing.js
```diff
@@ -0,0 +1,402 @@
...
+
+          assert.ok(!dispatchWrapper, 'should not have dispatch wrapper event')
+
+          done()
+        })
+    })
+  })
+
+  describe('error handler tracing', function () {
+    it('should trace error handlers (fn.length === 4)', function (done) {
```
Qard: Should this one capture the error even though it is considered handled overall?

Qard (approved):
Generally LGTM. I think we might want to validate if handled errors are still captured on the span prior to the handler though.

test/tracing.js
logaretm: Not entirely sure, APMs will usually have a global error handler or whatever to catch them so having them traced might be an overkill and may duplicate error emissions for those who don't dedup them.

WDYT?

Qard: But does it having a handler here result in the error _not_ being present on the span for the layer that lead to the error handler? Whatever the behaviour is, I think it would be useful to validate it more explicitly in this particular test. We should have a clear model of what errors happen where.

logaretm: Thanks! I added more tests to pin down various behaviors. Here is how errors flow right now:

**No error handler (no `app.use`):**

- Route throws sync -> `error` channel publishes for the route layer. Router responds 500.
- Route promise rejects -> `error` channel publishes for the route layer. Router responds 500.

**With error handler:**

- Route throws sync -> `error` channel publishes for the route, handler runs, nothing on the `error` channel for the handler.
- Route rejects async -> same, `error` channel publishes for the route, handler clean.

**With error handler that also throws/rejects:**

- Route throws sync, handler throws -> `error` channel publishes twice, once per layer. Two legitimately failed executions.
- Route rejects async, handler throws -> same.

For an APM: both throw cases produce two `error` events on the channel — one with `ctx.layer` pointing at the route, one at the error handler. Each event targets its own span, so the APM just marks both spans as failed.

In the case for `next(err)` then the child error handler span will be marked, not the parent one. Which is fine I suppose since it technically recovered as you've said.

Tests in 1ad1b9c.

test/tracing.js
```diff
@@ -211,6 +211,132 @@ describeTracing('TracingChannel', function () {
...
+          assert.ok(errorHandlerStart < errorHandlerEnd,
+            'error handler start should come before its own end')
+          assert.ok(errorHandlerEnd < failingEnd,
+            'error handler end should come before failing layer end — nesting contract: the error handler that runs via next(err) is nested inside the layer that triggered it, letting APMs attribute the error to the correct parent span')
+
+          done()
+        })
+    })
+
+    it('should not emit error on the originating layer when next(err) is unhandled', function (done) {
```
Qard: This seems problematic. Unhandled errors should definitely _always_ be observable.

Qard (commented):
Hmm...this handling seems a bit incorrect. If an error is unhandled it should _definitely_ be reported. And even if it _is_ handled we should still probably be reporting it so we can capture cases like a too permissive error handler silencing legitimate failures.

I think _probably_ what we should be doing is _always_ capturing the error and just adding some handled flag to the context to mark when it has been handled somehow. That way we can still report it with the detail that the application thinks it has been handled while still allowing human judgement to look at the span and see if that assessment actually makes sense.

test/tracing.js
logaretm: Okay agreed, this could've become a real gap and it's always easier to dedup rather than miss events.

I pushed a fix to that in 90ed7e2 and adjusted the tests to pin that behavior.

Qard (approved):
I think it's in a reasonable state now. Might be good to also get a review from @rochdev or @flarna for some more perspectives from other APM vendors. The specific error semantics I think might be a bit up to interpretation how they should be handled, but I _think_ this shape should generally provide the context that is needed for the different ways we think about route errors. 🤔

lib/layer.js
```diff
@@ -25,6 +26,20 @@ const deprecate = require('depd')('router')
 const TRAILING_SLASH_REGEXP = /\/+$/
 const MATCHING_GROUP_REGEXP = /\((?:\?<(.*?)>)?(?!\?)/g
 
+/**
+ * TracingChannel setup.
+ * @private
+ */
+
+const requestChannel = dc.tracingChannel && dc.tracingChannel('express:request')
```
Flarna: Why is `express:request` channel used here?
I think it would be better to use a dedicated channel for router to allow consumers to decide on what to receive and how to handle it.
I think router component can be also used without express.

logaretm: Good point, I don't think it has to be `express` prefixed but it needs a unique prefix for sure which is why I wanted some opinions on what it should be. My concern is `router:request` is too generic and can easily conflict with other names in the ecosystem.


I had this in the PR description:

> The channel name right now is express:request, in the original PR it was request but I think channel names need to have a relationship with the module name otherwise we risk duplication in the ecosystem.

WDYT? Maybe `pillarjs:router:request`?

Flarna: There seems to be a [guideline](https://nodejs.org/docs/latest/api/diagnostics_channel.html#tracingchannel-channels).
So I guess the string to use would be `pillarjs.router.request` as `tracing:` prefix and event name postfix is added by `dc.tracingChannel()`.

logaretm: Doesn't seem like there is a consistent standard, docs do suggest that but others just went with `lib:something:action` formats.

Anyways I have no strong opinions here and I think as long as the name is unique enough, it is arbitrary. I renamed it to your suggestion 👍

Flarna:
The actual data published by the tracing channel is the `Layer` instance, `req` and `res`  objects.

If I understand the router module correct `layer` is an internal object which gets exposed here. Is there any guarantee that the relevant fields on that object stay intact? In fact they become part of the public interface here.

Also I don't find in the docs here that `req.route`  is a public property. Is it safe for consumers to rely on that as it is seen part of public API and following Semver regarding changes?

In general I would love some section in the docs describing the diag channel and it's emitted data.

logaretm:
> Is it safe for consumers to rely on that as it is seen part of public API and following Semver regarding changes?

I can't answer that as it usually up to maintainers to decide this but I would say it does imply a public API contract and should be subject to smever.

As for the `req.route`, I'm not sure the AI created a snippet with it and it worked 🤷‍♂️ 

The `layer` object is strictly private today so this woulda actually expose it publicly.

> In general I would love some section in the docs describing the diag channel and it's emitted data.

I think we have 3 options here:

- Define a curated list of properties and expose those instead as public API.
- Layer objects become a public API.
- Continue to emit layer, but document a stable partial of it.

I think that would be up to the maintainers to decide, but all options work for us.

Qard:
I generally suggest that event contents be considered public API surface, but I would not say that is mandatory as it is very common that diagnostics_channel events would be observing what would otherwise be internal implementation details. I _would_ ask though that any exposure of internals be explicitly called out as internal to inform consumers that they should consume those events with defensive accesses to handle cases where shapes may change.

I don't fully think calling it public API surface is quite the right characterization of the intent.

lib/layer.js
bjohansebas: I think it should just be `router.request` and not `pillarjs.router.request`. I say this because pillarjs is more like the organization name and that could change, and we might move this package to the expressjs organization (not saying we’ll do it now, but it could happen, so we shouldn’t be tied to the GitHub organization name).

bjohansebas:
> Also I don't find in the docs here that req.route is a public property. Is it safe for consumers to rely on that as it is seen part of public API and following Semver regarding changes?

Yes, it’s part of the public API (see 
https://github.com/pillarjs/router/blob/cb2a5fcdf31b9cc92296183ceeff8126b5c57453/index.js#L281), so it should follow semver. That means anyone can depend on it. There might not be documentation (which I’ll add to my TODO once I’m back on the project, in case no one else has done it or if it doesn’t exist yet, since I don’t have time to verify it right now).

bjohansebas (commented):
Please also add documentation about this in the README.

logaretm:
@bjohansebas I added a section on diagnostic channels and their payload, and marked `layer` as internal impl.

lib/layer.js
bjohansebas: Okay, I understand the concern. In that case, instead of using pillarjs, it’s better to use express, since it’s officially under the expressjs project.

How does Fastify handle it?

logaretm: So should we go for `express.request` or `express.router.request`? To me its arbitrary but perhaps you have a strong opinion?

Fastify uses `fastify.request.handler` as [seen here](https://arc.net/l/quote/fibppzsf).

Flarna: I think `express.request` is a bit too generic. There might be other parts in express describing the request lifecycle where the router component is not included.
e.g. `express.request.router` or `express.router.request` - tending towards `express.request.router` like fastify and hierarchic ordering.

logaretm: Went with `express.router.request` since it describes the action better "Express's router is handling a request".

But happy to change to whatever that is acceptable to get this through.

lib/layer.js
```diff
@@ -25,6 +26,20 @@ const deprecate = require('depd')('router')
...
+ * @private
+ */
+
+const requestChannel = dc.tracingChannel && dc.tracingChannel('express.router.request')
+
+/**
+ * Check if the channel has subscribers.
+ */
+function shouldTrace (ch) {
+  return ch && ch.hasSubscribers !== false
```
bjohansebas: Since we support Node 18.0.0, users on those versions will get true, because (undefined !== false) === true.

https://nodejs.org/api/diagnostics_channel.html#diagnostics_channeltracingchannelnameorchannels

README.md
```diff
@@ -400,6 +400,56 @@ router.route('/pet/:id')
...
+loggers) hook into middleware and route handler execution without monkey-patching.
+
+Each layer's handler invocation publishes the standard tracing channel sub-events
+(`start`, `end`, `asyncStart`, `asyncEnd`, `error`). The published context object
+contains:
+
+- `req`: the incoming `http.IncomingMessage`
+- `res`: the `http.ServerResponse`
+- `layer`: the internal `Layer` instance being invoked (exposes `.name`, `.path`, `.handle`, etc.). Note that `Layer` is an internal implementation detail and its shape may change between releases.
+- `error`: the error passed to `next(err)`, when applicable
```
bjohansebas: If someone did something like this, I know it's a very edge case, but it's still something we should fix since it's allowed by the way the router is designed:

```js
router.get('/via-throw', function viaThrow (req, res, next) {
  throw 'route' // eslint-disable-line no-throw-literal
})
router.get('/via-throw', (req, res) => res.end('second route ran'))
```

the error event would still be emitted, so it wouldn't be limited to just `next(err)`. Even so, the request would still end up returning a `200`.

lib/layer.js
```diff
@@ -147,16 +151,52 @@ Layer.prototype.handleRequest = function handleRequest (req, res, next) {
...
+  const tracing = ctxFactory && shouldTrace(requestChannel)
+  const ctx = tracing ? ctxFactory() : null
+  const wrappedNext = tracing
+    ? function (err) {
+      // 'route' and 'router' are control-flow sentinels (skip route / exit router),
+      // not real errors — don't pollute the error channel with them.
+      if (err && err !== 'route' && err !== 'router') {
+        ctx.error = err
+        // Explicitly publish the error to the error channel
+        requestChannel.error.publish(ctx)
```
bjohansebas: Okay, we have a small bug here. When there are mounted sub-routers, if an error occurs, the parent router ends up publishing a duplicate error event. For example:

```js

const http = require('node:http')
const dc = require('node:diagnostics_channel')
const Router = require('/home/bjohansebas/Dev/oss/pillarjs/router')

const errorEvents = []

dc.tracingChannel('express.router.request').subscribe({
  error (ctx) {
    errorEvents.push({
      layer: ctx.layer && ctx.layer.name,
      handled: !!ctx.handled,
      message: ctx.error && ctx.error.message
    })
  }
})

function report (title) {
  console.log('\n%s', title)
  console.log('  error events published: %d %s', errorEvents.length,
    errorEvents.length === 1 ? '(OK)' : '(BUG: should be exactly 1, on the origin layer)')
  for (const e of errorEvents) {
    console.log('    layer=%s handled=%s message=%j', e.layer, e.handled, e.message)
  }
  errorEvents.length = 0
}

async function run (router, path) {
  const server = http.createServer(function (req, res) {
    router(req, res, function (err) {
      res.statusCode = err ? 500 : 404
      res.end(String(err || 'not found'))
    })
  })
  await new Promise(r => server.listen(0, r))
  await fetch('http://localhost:' + server.address().port + path)
  await new Promise(r => setImmediate(r))
  server.close()
}

async function main () {
  // Scenario A: mounted sub-router — a single next(err) in the inner handler,
  // recovered by an outer error handler.
  {
    const outer = new Router()
    const nested = new Router()

    nested.get('/bar', function innerHandler (req, res, next) {
      next(new Error('boom'))
    })

    outer.use('/foo', nested)

    outer.use(function recover (err, req, res, next) { // eslint-disable-line no-unused-vars
      res.statusCode = 500
      res.end(err.message)
    })

    await run(outer, '/foo/bar')
    report('Scenario A: nested router (1 mount level)')
  }

  // Scenario A2: two mount levels — the error is republished on EVERY ancestor.
  {
    const outer = new Router()
    const mid = new Router()
    const deep = new Router()

    deep.get('/baz', function deepHandler (req, res, next) {
      next(new Error('boom'))
    })

    mid.use('/bar', deep)
    outer.use('/foo', mid)

    outer.use(function recover (err, req, res, next) { // eslint-disable-line no-unused-vars
      res.statusCode = 500
      res.end(err.message)
    })

    await run(outer, '/foo/bar/baz')
    report('Scenario A2: nested router (2 mount levels)')
  }

  // Scenario B: error handler that forwards the error with next(err).
  {
    const router = new Router()

    router.get('/fail', function origin (req, res, next) {
      next(new Error('boom'))
    })

    router.use(function forwardingHandler (err, req, res, next) {
      next(err) // does not handle it, delegates to the next one
    })

    router.use(function recover (err, req, res, next) { // eslint-disable-line no-unused-vars
      res.statusCode = 500
      res.end(err.message)
    })

    await run(router, '/fail')
    report('Scenario B: error handler forwarding with next(err)')
  }
}

main()

```

In principle, the error event should only be published once.

lib/layer.js
```diff
@@ -147,16 +151,52 @@ Layer.prototype.handleRequest = function handleRequest (req, res, next) {
...
       }
+      next(err)
+    }
+    : next
 
+  try {
+    const ret = tracing
+      ? requestChannel.tracePromise(function () { return handlePromise(exec(wrappedNext)) }, ctx)
```
bjohansebas: If someone does this, or manages to trigger it somehow:

```js
router.get('/reject', async function rejectingHandler (req, res) {
  throw undefined // same as: return Promise.reject()
})
```

the error emitted in the event would be `undefined`, when it should instead be `Error: Rejected promise`.

lib/layer.js
```diff
@@ -147,16 +151,52 @@ Layer.prototype.handleRequest = function handleRequest (req, res, next) {
...
+      return fn(req, res, wrappedNext)
+    }, null, next)
+  }
+
+  const layer = this
+  invokeWithTrace(function (wrappedNext) {
+    return fn(req, res, wrappedNext)
+  }, function () {
+    return { req, res, layer }
+  }, next)
```
bjohansebas: This avoids allocating objects on every request when the tracing channel isn't enabled, and also improves readability.

```suggestion
if (this.route || !shouldTrace(requestChannel)) {
    try {
      // invoke function
      const ret = handlePromise(fn(req, res, next))

      // wait for returned promise
      if (ret) {
        ret.then(null, function (error) {
          next(error || new Error('Rejected promise'))
        })
      }
    } catch (err) {
      next(err)
    }
    return
```

lib/layer.js
```diff
@@ -111,23 +126,12 @@ Layer.prototype.handleError = function handleError (error, req, res, next) {
...
-    }
-  } catch (err) {
-    next(err)
-  }
+  const layer = this
+  invokeWithTrace(function (wrappedNext) {
+    return fn(error, req, res, wrappedNext)
+  }, function () {
+    return { req, res, layer, error, handled: true }
+  }, next)
```
bjohansebas: This avoids creating allocations per request when no tracing channel is enabled, keeping the behavior consistent with what's already in `master`.

```suggestion
  if (!shouldTrace(requestChannel)) {
    try {
      // invoke function
      const ret = handlePromise(fn(error, req, res, next))

      // wait for returned promise
      if (ret) {
        ret.then(null, function (error) {
          next(error || new Error('Rejected promise'))
        })
      }
    } catch (err) {
      next(err)
    }
    return
  }

  const layer = this
  invokeWithTrace(function (wrappedNext) {
    return fn(error, req, res, wrappedNext)
  }, { req, res, layer, error, handled: true }, next)
```

lib/layer.js
bjohansebas: To simplify things when you implement the other changes.
```suggestion
function invokeWithTrace (exec, ctx, next) {
  const wrappedNext = function (err) {
    // 'route' and 'router' are control-flow sentinels (skip route / exit router),
    // not real errors — don't pollute the error channel with them.
    if (err && err !== 'route' && err !== 'router') {
      ctx.error = err
      // Explicitly publish the error to the error channel
      requestChannel.error.publish(ctx)
    }
    next(err)
  }

  try {
    const ret = requestChannel.tracePromise(function () { return handlePromise(exec(wrappedNext)) }, ctx)

```

bjohansebas (changes requested):
Good job! I noticed a few details, and I'll open a PR to update the documentation on the website

edit: Note that the refactors I mentioned don't fix the bugs I found.

lib/layer.js
logaretm: Yep this is a trade off I suppose. I think we should then check for all channels own subscribers in that case.

lib/layer.js
logaretm: That's a good catch. I will dedup the errors published in relation to the request object.

lib/layer.js
logaretm: Usually this is fine, but since it gets normalized later it the channel error should match. Fixing this.

README.md
logaretm: I added an exception for these routing signals, should be ignored now.

lib/layer.js
```diff
@@ -147,25 +174,131 @@ Layer.prototype.handleRequest = function handleRequest (req, res, next) {
...
+        recordError(ctx.req, error)
+        next(error)
       })
     }
   } catch (err) {
+    // tracePromise already reported the sync throw on this layer; record it so
+    // outer layers don't report it again.
+    recordError(ctx.req, err)
     next(err)
   }
```
bjohansebas: Okay, I think we can simplify this further while also fixing a few other cases where I found differences compared to `master`. I've described those cases in the comments, and you can take a look and decide whether you think this is the best approach.

```suggestion
 requestChannel.tracePromise(function () {
    let out
    try {
      out = handlePromise(exec(wrappedNext))
    } catch (err) {
      // forwarded verbatim: a falsy throw keeps routing, like the untraced path
      wrappedNext(err)
      return
    }

    if (out) {
      return out.then(undefined, function (err) {
        wrappedNext(err || new Error('Rejected promise'))
      })
    }
  }, ctx)
```

bjohansebas: For example, when a falsy value is thrown in synchronous mode, like in the following example:

 ```js
 dc.tracingChannel('express.router.request').subscribe(handlers)

 router.get('/falsy', function throwsFalsy(req, res) {
   throw undefined // eslint-disable-line no-throw-literal
 })

 router.get('/falsy', function continues(req, res) {
   res.statusCode = 200
   res.end('continued')
 })

 request(server)
   .get('/falsy')
   .expect(200, 'continued', function (err) {
     if (err) return done(err)

     const errorEvents = events.filter(function (e) { return e.phase === 'error' })
     assert.equal(errorEvents.length, 0,
       'a falsy throw is treated as "no error" by the router — subscribing must not divert it to error handling')

     done()
   })
 ```

Without tracing enabled, the router doesn't treat this as an error and simply continues as normal. However, once tracing is enabled, it now automatically treats it as an error. It's also interesting that we don't currently have a test covering the normal behavior. It's definitely an edge case, but enabling tracing shouldn't change the router's behavior.

Could you add a test for this case 
```js
  it('should continue routing after throwing a falsy value', function (done) {
        const router = new Router()
        const server = createServer(router)

        router.use(function handle (req, res, next) {
          throw undefined // eslint-disable-line no-throw-literal
        })

        router.use(sawError)

        router.use(helloWorld)

        request(server)
          .get('/')
          .expect(200, 'hello, world', done)
      })
``

Comment (bjohansebas) [reply lib/layer.js]: Another case we have is when the same error is rethrown. It's technically valid—people do some crazy things 😄—but here's the scenario:

 ```js
 dc.tracingChannel('express.router.request').subscribe(handlers)

 router.get('/fail', function origin (req, res, next) {
   next(new Error('boom'))
 })
 router.use(function rethrowing (err, req, res, next) { // eslint-disable-line no-unused-vars
   throw err
 })
 router.use(function recover (err, req, res, next) { // eslint-disable-line no-unused-vars
   res.statusCode = 500
   res.end(err.message)
 })
 request(server)
   .get('/fail')
   .expect(500, 'boom', function (err) {
     if (err) return done(err)

     const errorEvents = events.filter(function (e) { return e.phase === 'error' })
     assert.equal(errorEvents.length, 1,
       'rethrowing the same error must not re-report it — throw and next(err) are equivalent to the router')
     assert.equal(errorEvents[0].ctx.layer.name, 'origin')
     done()
   })
 ```

In this case, it currently emits the error event twice. I don't think that makes sense—we shouldn't report the same error twice. From the router's perspective, `throw err` and `next(err)` are equivalent, so a single error should only produce a single error event.

It all depends on the user's coding style.

File: test/tracing.js [hunk 11]
```diff
@@ -0,0 +1,981 @@
...
+      start (ctx) { events.push({ phase: 'start', ctx }) },
+      end (ctx) { events.push({ phase: 'end', ctx }) },
+      asyncStart (ctx) { events.push({ phase: 'asyncStart', ctx }) },
+      asyncEnd (ctx) { events.push({ phase: 'asyncEnd', ctx }) },
+      error (ctx) { events.push({ phase: 'error', ctx }) }
+    }
+  })
+
+  afterEach(function () {
+    dc.tracingChannel('express.router.request').unsubscribe(handlers)
```
bjohansebas: We use `'express.router.request'` a lot. We could probably define it as a constant so we don't have a magic string repeated throughout the tests

test/tracing.js
```diff
@@ -0,0 +1,981 @@
...
+        res.end(err.message)
+      })
+
+      request(server)
+        .get('/fail')
+        .expect(500, 'boom', function (err) {
+          if (err) return done(err)
+
+          const byLayer = function (name) {
+            return function (e) { return e.ctx.layer && e.ctx.layer.name === name }
```
bjohansebas: This could be a test helper. We could move it into a function at the bottom of the file and call it wherever it's needed, so we don't have to repeat the same function in multiple places

lib/layer.js
logaretm: Yep that looks much better, I enumerated all cases again and ran what I just pushed against them, should be good now 👍 

Thanks for double checking here!

README.md
```diff
@@ -400,6 +400,58 @@ router.route('/pet/:id')
...
+
+Each layer's handler invocation publishes the standard tracing channel sub-events
+(`start`, `end`, `asyncStart`, `asyncEnd`, `error`). The published context object
+contains:
+
+- `req`: the incoming `http.IncomingMessage`
+- `res`: the `http.ServerResponse`
+- `layer`: the internal `Layer` instance being invoked (exposes `.name`, `.path`, `.handle`, etc.). Note that `Layer` is an internal implementation detail and its shape may change between releases.
+- `error`: the error the layer failed with, when applicable
+- `handled`: `true` when the layer is an error-handling middleware (4-arg signature)
```
bjohansebas: I think we should rename it to something like `errorHandler`, maybe?

```suggestion
- `errorHandled`: `true` when the layer is an error-handling middleware (4-arg signature)
```

bjohansebas (approved):
LGTM. It looks like `tracePromise` cuts performance roughly in half when tracing is enabled. I'm not sure if that's expected behavior for an APM. We could probably improve performance a bit by using `.start.runStores`, although that would be semantically different.

README.md
logaretm: That's fair, the clearer the better I suppose. Maybe `isErrorHandler` or `errorHandler` actually? since it doesn't mean that the error is handled, just that the handler is an error handler.

logaretm:
> We could probably improve performance a bit by using .start.runStores, although that would be semantically different.

Yea that would mean it doesn't follow the tracing channels regular flow which can be important for APMs since it has to be predictable. `tracePromise` should essentially be a syntax sugar over start.runStores.

Let me see if we can benchmark this.

bjohansebas:
I'm testing on my local machine, which isn't the most stable environment for benchmarks, but I couldn't find much else to optimize. I also think Fastify and others pay the same cost when using tracePromise, so it seems more like a Node.js overhead than something we can really improve on our side

logaretm:
So I did a test and you are right it is 1.9x faster for me, but only on the sync handlers case. Async handlers are more or less the same.

We could do what we did with GraphQL and strap the whole 5 events with runStores and manual publishes. Let me see if that gives us a win in sync case and passes the context propagation tests, it should.

EDIT: Just did, for sync handlers it will effectively follow traceSync behavior, for non async handlers it will follow the same semantics as `tracePromise`. This regains 1.9x speed for those sync handlers on my machine.

bjohansebas:
SGTM. @Qard @Flarna , since you have more context on this as well, do you think this is a good approach?

logaretm:
@UlisesGascon any chance we can get this merged? many involved folks already signed off.

PR #197, opened 2026-05-09
feat: add typescript types

gameroman:


bjohansebas:
Hey, could you add the TypeScript test infrastructure we added in on-finished? Sorry for the delay. https://github.com/jshttp/on-finished/pull/69

test/types.ts
```diff
@@ -0,0 +1,25 @@
...
+});
+
+router.get("/", function (req, res, next) {
+  expectTypeOf(req).toEqualTypeOf<IncomingMessage>();
+  expectTypeOf(res).toEqualTypeOf<ServerResponse>();
+  expectTypeOf(next).toBeFunction();
+});
+
+// typescript has no way to infer the types automatically
+router.get<ErrorRequestHandler>("/", function (err, req, res, next) {
```
gameroman, line 23: Basically, typescript can not automatically infer that the callback type is `ErrorRequestHandler` so if you don't provide the type param, it'll just infer it's params as `any`

The only way to solve this is to change everything completely, changing callback's param to an object like `{ err, req, res, next }`, instead of 4 different params, this is what `hono`, `elysiajs` and other do

This would obviously require a new major so I haven't touched anything there

So just letting know that the current logic cannot be fully expressed in typescript without manually providing the type parameter on the user's side

lib/route.d.ts
```diff
@@ -0,0 +1,37 @@
+import type { IncomingMessage, ServerResponse } from 'node:http'
+
+export type NextFunction = (err?: any) => void
+
+export type RequestHandler = (req: IncomingMessage, res: ServerResponse, next: NextFunction) => void
+export type ErrorRequestHandler = (err: Error, req: IncomingMessage, res: ServerResponse, next: NextFunction) => void
+
+export type Handler = RequestHandler | ErrorRequestHandler
+
+export type RouteMethod<T> = <F extends Handler = RequestHandler>(handler: F) => T
```
bjohansebas:
```suggestion
export type HandlerArg<F extends Handler = Handler> = F | ReadonlyArray<HandlerArg<F>>

export type RouteMethod<T> = <F extends Handler = RequestHandler>(handler: HandlerArg<F>, ...handlers: Array<HandlerArg<F>>) => T

```

To handle cases like:

```js
router.get('/', [fn1, fn2])
```

index.d.ts
```diff
@@ -0,0 +1,53 @@
+import type { IncomingMessage, ServerResponse } from 'node:http'
+import type {
+  Handler,
+  NextFunction,
+  RequestHandler,
+  Route as IRoute
+} from './lib/route'
+import Route = require('./lib/route')
+
+export type RouterMethod<T> = <F extends Handler = RequestHandler>(path: Path, handler: F) => T
```
bjohansebas:
```suggestion
export type RouterMethod<T> = <F extends Handler = RequestHandler>(path: Path, handler: HandlerArg<F>, ...handlers: Array<HandlerArg<F>>) => T

```

For when you import `handlerArgs`, so these get the fix as well

index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+  mergeParams: boolean
+  strict: boolean
+
+  handle(req: IncomingMessage, res: ServerResponse, callback: NextFunction): void
+  param(name: string, fn: (req: any, res: any, next: NextFunction, value: any, name: string) => void): this
+  route(path: Path): IRoute
+
+  use: {
+    <F extends Handler = RequestHandler>(path: Path, handler: F): Router
+    <F extends Handler = RequestHandler>(handler: F): Router
```
bjohansebas:
```suggestion
   <F extends Handler = RequestHandler>(path: Path, handler: HandlerArg<F>, ...handlers: Array<HandlerArg<F>>): Router
    <F extends Handler = RequestHandler>(handler: HandlerArg<F>, ...handlers: Array<HandlerArg<F>>): Router
```

index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+interface RouterConstructor {
+  (options?: RouterOptions): Router
+  new (options?: RouterOptions): Router
+  Route: typeof Route
+}
+
+declare const router: RouterConstructor
+export = router
```
bjohansebas, line 56: We're mixing `export =` with other exports, which will cause issues for anyone using `skipLibCheck: false`. `@tsconfig/node18` enables `skipLibCheck` by default, which is why the CI doesn't catch this error. If you need help figuring out how to fix it, let me know.

this applies to both files.

lib/route.d.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  all: RouteMethod<this>
+  get: RouteMethod<this>
+  post: RouteMethod<this>
+  put: RouteMethod<this>
+  delete: RouteMethod<this>
+  patch: RouteMethod<this>
+  options: RouteMethod<this>
+  head: RouteMethod<this>
+}
```
bjohansebas: This makes it easier to maintain. The only thing I don't like is that we have to maintain a manual list for each method, but that's basically a limitation of Node.js's own types. And yes, I know `query` isn't available in every version, but we support it anyway, so that's a trade-off we have to accept.

```suggestion
export type Method =
  | 'acl' | 'bind' | 'checkout' | 'connect' | 'copy' | 'delete' | 'get'
  | 'head' | 'link' | 'lock' | 'm-search' | 'merge' | 'mkactivity'
  | 'mkcalendar' | 'mkcol' | 'move' | 'notify' | 'options' | 'patch'
  | 'post' | 'propfind' | 'proppatch' | 'purge' | 'put' | 'query'
  | 'rebind' | 'report' | 'search' | 'source' | 'subscribe' | 'trace'
  | 'unbind' | 'unlink' | 'unlock' | 'unsubscribe'

export type MethodHandlers<T> = { [M in Method | 'all']: RouteMethod<T> }

export interface Route extends MethodHandlers<Route> {
  path: string
  stack: any[]
  methods: Record<string, boolean>

  _handlesMethod(method: string): boolean
  _methods(): string[]
  dispatch(req: IncomingMessage, res: ServerResponse, done: NextFunction): void
}
```

index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type Path = string | RegExp | Array<string | RegExp>
+
+interface RouterOptions {
+  caseSensitive?: boolean
+  mergeParams?: boolean
+  strict?: boolean
+}
+
+interface Router extends RequestHandler {
```
bjohansebas: To implement the method changes and export it as well.

```suggestion
type MethodHandlers<T> = { [M in Method | 'all']: RouterMethod<T> }

export interface Router extends RequestHandler, MethodHandlers<Router> {
```

index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+  RequestHandler,
+  Route as IRoute
+} from './lib/route'
+import Route = require('./lib/route')
+
+export type RouterMethod<T> = <F extends Handler = RequestHandler>(path: Path, handler: F) => T
+
+export type Path = string | RegExp | Array<string | RegExp>
+
+interface RouterOptions {
```
bjohansebas: It's a good idea to export it.

```suggestion
export interface RouterOptions {
```

lib/route.d.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type RequestHandler = (req: IncomingMessage, res: ServerResponse, next: NextFunction) => void
+export type ErrorRequestHandler = (err: Error, req: IncomingMessage, res: ServerResponse, next: NextFunction) => void
+
+export type Handler = RequestHandler | ErrorRequestHandler
+
+export type RouteMethod<T> = <F extends Handler = RequestHandler>(handler: F) => T
+
+export interface Route {
+  path: string
```
bjohansebas, line 26: I'd like us to use the `Path` type from `index.d.ts` here as well, but unfortunately we'd end up breaking more than we'd gain, since it's been this way in the Express types for years.

lib/route.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  put: RouteMethod<this>
+  delete: RouteMethod<this>
+  patch: RouteMethod<this>
+  options: RouteMethod<this>
+  head: RouteMethod<this>
+}
+
+interface RouteConstructor {
+  (path: string): Route
+  new (path: string): Route
```
bjohansebas: It's always `constructor` :)

```suggestion
  new (path: Path): Route
```

bjohansebas: Note that this uses the `Path` type from `index.d.ts`, which is what we should be using here as well

test/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  expectTypeOf(err).toEqualTypeOf<Error>();
+  expectTypeOf(req).toEqualTypeOf<IncomingMessage>();
+  expectTypeOf(res).toEqualTypeOf<ServerResponse>();
+  expectTypeOf(next).toBeFunction();
+});
```
bjohansebas, line 28: Okay, these are a few more valid cases that the types should support.

```suggestion
import { IncomingMessage, ServerResponse } from "node:http";
import { expectTypeOf } from "expect-type";
import Router, { ErrorRequestHandler, Method, NextFunction, Path, RequestHandler } from "..";

const router = Router();

const handler: RequestHandler = (req, res, next) => next();
const other: RequestHandler = (req, res, next) => next();

// handler parameters are inferred
router.get("/", function (req, res) {
  expectTypeOf(req).toEqualTypeOf<IncomingMessage>();
  expectTypeOf(res).toEqualTypeOf<ServerResponse>();
});

router.get("/", function (req, res, next) {
  expectTypeOf(req).toEqualTypeOf<IncomingMessage>();
  expectTypeOf(res).toEqualTypeOf<ServerResponse>();
  expectTypeOf(next).toBeFunction();
});

// typescript has no way to infer the types automatically
router.get<ErrorRequestHandler>("/", function (err, req, res, next) {
  expectTypeOf(err).toEqualTypeOf<Error>();
  expectTypeOf(req).toEqualTypeOf<IncomingMessage>();
  expectTypeOf(res).toEqualTypeOf<ServerResponse>();
  expectTypeOf(next).toBeFunction();
});

// methods accept multiple handlers and nested arrays (flattened at runtime)
router.get("/", handler, other);
router.get("/", [handler, other]);
router.get("/", [handler, [other]], handler);

// at least one handler is required (runtime throws TypeError)
// @ts-expect-error
router.get("/");
// handlers must be functions or arrays of them
// @ts-expect-error
router.get("/", "not-a-handler");

// paths may be a string, RegExp, or array of either
router.get(/^\/x/, handler);
router.get(["/a", /^\/b/], handler);
// @ts-expect-error
router.get(42, handler);

// use: with or without path, multiple handlers, arrays
router.use(handler);
router.use(handler, other);
router.use([handler, [other]]);
router.use("/a", handler, [other], handler);
router.use(["/a", "/b"], handler);
router.use(/^\/c/, handler);
// @ts-expect-error path without any handler (runtime throws TypeError)
router.use("/x");
// @ts-expect-error
router.use();

// verb methods and use return the router for chaining
expectTypeOf(router.get("/", handler)).toEqualTypeOf(router);
router.use(handler).use("/a", other).get("/", handler);

// one method per node:http METHODS entry, not just the classic 8
router.trace("/", handler);
router.search("/", handler);
router.propfind("/", handler);
router["m-search"]("/", handler);
// @ts-expect-error not an HTTP method
router.foo("/", handler);

// error handlers are distinguished by the explicit generic
router.use<ErrorRequestHandler>(function (err, req, res, next) {
  expectTypeOf(err).toEqualTypeOf<Error>();
  next(err);
});

// param callbacks receive typed req/res
router.param("id", function (req, res, next, value, name) {
  expectTypeOf(req).toEqualTypeOf<IncomingMessage>();
  expectTypeOf(res).toEqualTypeOf<ServerResponse>();
  expectTypeOf(next).toEqualTypeOf<NextFunction>();
  expectTypeOf(name).toBeString();
});

// router is itself a request handler (mountable in a server or another router)
expectTypeOf(router).toExtend<RequestHandler>();
router.use(Router());

// the instance, options, and helper types are nameable by consumers
const mounted: Router.Router = Router({ strict: true, caseSensitive: true, mergeParams: true });
mounted.use(handler);
const opts: Router.RouterOptions = { strict: true };
Router(opts);
new Router(opts);
// @ts-expect-error unknown options are rejected
Router({ nope: true });
expectTypeOf<Method>().toExtend<string>();
expectTypeOf<Path>().toEqualTypeOf<string | RegExp | Array<string | RegExp>>();

// routes: all verbs, multiple handlers, chaining
const route = router.route("/things/:id");
expectTypeOf(route.path).toBeString();
route.all(handler).get(handler, other).post([handler, [other]]);
route.checkout(handler);
route["m-search"](handler);
// @ts-expect-error at least one handler is required
route.put();

// Route must be constructed with new, and accepts any Path
new Router.Route("/x");
new Router.Route(/^\/x/);
new Router.Route(["/a", /^\/b/]);
// @ts-expect-error calling without new does not construct (strict mode)
Router.Route("/x");
// @ts-expect-error
new Router.Route(42);

```

bjohansebas (commented):
Good start! I found a few nits while doing the code review

PR #202, opened 2026-06-05
cleanup: replace `is-promise` with `instanceof Promise` check

### LINKED ISSUE #136: Consider dropping `is-promise` as a dependency

maksim-romanov:
Native Promises are universally supported, so the `is-promise` dependency
can be replaced with a strict `instanceof Promise` check. The deprecation
block added in #154 becomes unreachable but is left in place so this PR
stays minimal — it can be removed in a follow-up.

Closes #136


krzysdz:
Duplicate of #137

